PR #5, opened 2018-03-02
Added a potential command line running capability

blen-desta:
Got the idea from Mike. We can use this cli library to parse arguments passed in and uses defaults if arguments aren't provided. Not sure if this is the ideal place to put it. Just wanted to put it out here to discuss.

### DIFF
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -44,6 +44,11 @@
       <artifactId>suites</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>us.jimschubert</groupId>
+      <artifactId>kopper-typed</artifactId>
+      <version>0.0.3</version>
+    </dependency>
   </dependencies>
   
   <build>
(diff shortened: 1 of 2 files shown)


coyotesqrl:
If I'm going to use a cli parser, I always opt for https://github.com/spullara/cli-parser

😉

coyotesqrl:
Parser preference aside, I am definitely in favor of the change.

blen-desta:
I found this parser from https://github.com/KotlinBy/awesome-kotlin. Mike linked me https://github.com/Kotlin/kotlinx.cli but I couldn't find it in maven central.

blen-desta:
We were looking at kotlin-bases libraries

coyotesqrl:
That's reasonable. Stick with a Kotlin library.

blen-desta:
Do you have a preference of where to put it? I just have it in `distribution/src/main/kotlin`

coyotesqrl:
That seems very reasonable.

blen-desta:
Okay I'm gonna merge it and we can keep improving it

PR #8, opened 2018-03-08
Created new object for plugin return

bakejeyner:
Creates the initial object for plugin return type. The next step would be to actually return this object and refactor the code in order to
1. Accept this new object
2. Use the new properties of the object to reduce complexity

### DIFF
--- a/plugins/plugins-api/pom.xml
+++ b/plugins/plugins-api/pom.xml
@@ -1,16 +1,36 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0"
-         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-    <modelVersion>4.0.0</modelVersion>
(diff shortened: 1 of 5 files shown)


plugins/plugins-api/pom.xml
```diff
@@ -4,7 +4,15 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
-    <parent>
+  <dependencies>
```
blen-desta: Could you put dependencies under parent. Not a huge deal but we should be consistent.

plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpPostResponse.java
```diff
@@ -0,0 +1,144 @@
...
+ *       .samlForm(exampleSamlForm);
+ *   return idpPostResponseBuilder.build();
+ * </pre>
+ *
+ * </blockquote>
+ */
+public class IdpPostResponse {
+  private IdpPostResponse() {}
+
+  public static class IdpPostResponseBuilder {
```
coyotesqrl: ✏️ You can just name this `Builder`. Same for the sibling class. Then they're accessed like `new IdpPostResponse.Builder()`.

PR #9, opened 2018-03-14
Added object return to plugin methods and propagated object to verifiers

bakejeyner:
This PR piggy backs off of the last PR. It adds the previously made object to the plugin method's return type. The code is also refactored in the verification to accept this change.

### DIFF
--- a/plugins/ddf-plugins/src/main/kotlin/org/codice/compliance/saml/plugin/ddf/IdpResponderProvider.kt
+++ b/plugins/ddf-plugins/src/main/kotlin/org/codice/compliance/saml/plugin/ddf/IdpResponderProvider.kt
@@ -22,6 +22,8 @@ import org.apache.commons.lang3.StringUtils
 import org.apache.cxf.rs.security.saml.sso.SSOConstants.RELAY_STATE
 import org.apache.cxf.rs.security.saml.sso.SSOConstants.SAML_RESPONSE
 import org.codice.compliance.Common
+import org.codice.compliance.saml.plugin.IdpPostResponse
+import org.codice.compliance.saml.plugin.IdpRedirectResponse
 import org.codice.compliance.saml.plugin.IdpResponder
 import org.codice.security.saml.SamlProtocol
 import org.kohsuke.MetaInfServices
@@ -33,33 +35,16 @@ class IdpResponderProvider : IdpResponder {
         val REDIR_REGEX = """var encoded\s*=\s*"(.*)";""".toRegex()
     }
 
-    override fun getIdpRedirectResponse(originalResponse: Response): String? {
+    override fun getIdpRedirectResponse(originalResponse: Response): IdpRedirectResponse {
         val response = parseResponseAndSendRequest(originalResponse)
 
         /*
         * TODO "TODO "Manually change DDF IdP to respond with 302/303 status code for Redirect"
         * When ticket is finished, replace below url building line with:
         * .url(response.getHeader("Location"));
+        *
+        * And delete the html block comment
          */
-
-        /*
-        * TODO "Revamp pluggable portion return an object (made by us)"
-        * When tickets are completed, insert this code and delete all the code below it in this method.
-
-        val script = response
-                .then()
-                .extract()
-                .htmlPath()
-                .getNode("html")
-                .getNode("head")
-                .getNode("script")
-                .value()
-
-        val idpRedirectResponseBuilder = IdpRedirectResponse.Builder()
-        idpRedirectResponseBuilder.httpStatusCode(response.statusCode)
-            .url(REDIR_REGEX.find(script)?.groups?.get(1)?.value); //validate that this gives you the correct url
-        return idpRedirectResponseBuilder.build()
-        */
 
         /*************************
          * <html>
@@ -87,10 +72,13 @@ class IdpResponderProvider : IdpResponder {
                 .getNode("script")
                 .value()
 
-        return REDIR_REGEX.find(script)?.groups?.get(1)?.value
+        return IdpRedirectResponse.Builder().apply {
+            httpStatusCode(response.statusCode)
+            url(REDIR_REGEX.find(script)?.groups?.get(1)?.value)
+        }.build()
     }
 
-    override fun getIdpPostResponse(originalResponse: Response): String? {
+    override fun getIdpPostResponse(originalResponse: Response): IdpPostResponse {
         val response = parseResponseAndSendRequest(originalResponse)
 
         /*************************
@@ -104,42 +92,16 @@ class IdpResponderProvider : IdpResponder {
          * ...
          ************************/
 
-        /*
-        * TODO "Revamp pluggable portion return an object (made by us)"
-        * When ticket is completed, insert this code and delete all the code below it in this method.
-
-        val response = parseResponseAndSendRequest(originalResponse)
-        val idpPostResponseBuilder = IdpPostResponse.Builder()
-        idpPostResponseBuilder.httpStatusCode(response.statusCode)
-            .samlForm(response
-                .then()
-                .extract()
-                .htmlPath()
-                .getNode("html")
-                .getNode("body")
-                .getNode("form"));
-        return idpPostResponseBuilder.build()
-        */
-
-        val form = response
-                .then()
-                .extract()
-                .htmlPath()
-                .getNode("html")
-                .getNode("body")
-                .getNode("form")
-
-        val samlResponse = form
-                .getNodes("input")[1]
-                .getAttribute("value")
-
-        val relayState = form
-                .getNodes("input")[2]
-                .getAttribute("value")
-
-        return if (StringUtils.isNoneBlank(relayState))
-            "$RELAY_STATE=$relayState&$SAML_RESPONSE=$samlResponse"
-        else "$SAML_RESPONSE=$samlResponse"
+        return IdpPostResponse.Builder().apply {
+            httpStatusCode(response.statusCode)
+            samlForm(response
+                    .then()
+                    .extract()
+                    .htmlPath()
+                    .getNode("html")
+                    .getNode("body")
+                    .getNode("form"))
+        }.build()
     }
 
     /**
--- a/plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpPostResponse.java
+++ b/plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpPostResponse.java
@@ -13,6 +13,7 @@
  */
 package org.codice.compliance.saml.plugin;
 
+import static org.apache.commons.lang3.StringUtils.isNotEmpty;
 import static org.apache.cxf.rs.security.saml.sso.SSOConstants.RELAY_STATE;
 import static org.apache.cxf.rs.security.saml.sso.SSOConstants.SAML_RESPONSE;
 
@@ -47,7 +48,8 @@ import com.jayway.restassured.path.xml.element.Node;
  *
  * </blockquote>
  */
-public class IdpPostResponse {
+public class IdpPostResponse extends IdpResponse {
+
   private IdpPostResponse() {}
 
   public static class Builder {
@@ -70,15 +72,14 @@ public class IdpPostResponse {
     }
   }
 
-  // General
-  private int httpStatusCode;
-  private Node samlResponseForm;
-  private String samlResponse;
-  private String relayState;
+  private static final String VALUE = "value";
+  private static final String TYPE = "type";
+  private static final String NAME = "name";
 
-  // Flags
+  private Node samlResponseForm;
+
+  private boolean isSamlResponseHidden;
   private boolean isRelayStateHidden;
-  private boolean isRelayStateGiven;
 
   @SuppressWarnings("squid:S3398" /* Method in here to simplify builder class */)
   private void parseAndSetFormValues(Node samlResponseForm) {
@@ -86,59 +87,77 @@ public class IdpPostResponse {
 
     // Bindings 3.5.4 "If the message is a SAML response, then the form control MUST be named
     // SAMLResponse."
-    samlResponse =
+    Node samlResponseNode =
         samlResponseForm
-            .getNodes("input")
+            .children()
+            .list()
             .stream()
-            .filter(node -> node.name().equals(SAML_RESPONSE))
-            .map(Node::value)
-            .findFirst()
-            .orElse(null);
-
-    Node relayStateNode =
-        samlResponseForm
-            .getNodes("input")
-            .stream()
-            .filter(node -> node.name().equals(RELAY_STATE))
+            .filter(node -> SAML_RESPONSE.equals(node.attributes().get(NAME)))
             .findFirst()
             .orElse(null);
 
     // Bindings 3.5.4 "If a “RelayState” value is to accompany the SAML protocol message, it MUST be
-    // placed in an additional hidden form control named RelayState within the same form with the
-    // SAML message"
-    if (relayStateNode != null) {
-      relayState = relayStateNode.value();
-      if (relayStateNode.getAttribute("type") != null) {
-        isRelayStateHidden = relayStateNode.getAttribute("type").equals("hidden");
+    // placed in an additional **hidden** form control named RelayState within the same form with
+    // the SAML message"
+
+    Node relayStateNode =
+        samlResponseForm
+            .children()
+            .list()
+            .stream()
+            .filter(node -> RELAY_STATE.equals(node.attributes().get(NAME)))
+            .findFirst()
+            .orElse(null);
+
+    /*
+     * Bindings 3.5.4 "A SAML protocol message is form-encoded by... placing the result **in** a
+     * **hidden** form control within a form as defined by [HTML401] Section 17"
+     *
+     * The two key words here are "in" and "hidden"
+     *
+     * Assuming "in" in the above quote means in either the value attribute or in the value
+     * itself.
+     *
+     * And "hidden" means both the SAMLResponse and RelayState MUST be placed in "hidden" form controls
+     */
+    // SAMLResponse portion
+    if (samlResponseNode != null) {
+
+      if (isNotEmpty(samlResponseNode.value())) {
+        samlResponse = samlResponseNode.value();
+      } else if (isNotEmpty(samlResponseNode.attributes().get(VALUE))) {
+        samlResponse = samlResponseNode.attributes().get(VALUE);
+      }
+
+      if (isNotEmpty(samlResponseNode.getAttribute(TYPE))) {
+        isSamlResponseHidden = samlResponseNode.getAttribute(TYPE).equals("hidden");
       }
     }
-  }
 
-  public int getHttpStatusCode() {
-    return httpStatusCode;
-  }
+    // RelayState portion
+    if (relayStateNode != null) {
 
-  public String getSamlResponse() {
-    return samlResponse;
-  }
+      if (isNotEmpty(relayStateNode.value())) {
+        relayState = relayStateNode.value();
+      } else if (isNotEmpty(relayStateNode.attributes().get(VALUE))) {
+        relayState = relayStateNode.attributes().get(VALUE);
+      }
 
-  public String getRelayState() {
-    return relayState;
+      if (isNotEmpty(relayStateNode.getAttribute(TYPE))) {
+        isRelayStateHidden = relayStateNode.getAttribute(TYPE).equals("hidden");
+      }
+    }
   }
 
   public Node getSamlResponseForm() {
     return samlResponseForm;
   }
 
+  public boolean isSamlResponseHidden() {
+    return isSamlResponseHidden;
+  }
+
   public boolean isRelayStateHidden() {
     return isRelayStateHidden;
-  }
-
-  public boolean isRelayStateGiven() {
-    return isRelayStateGiven;
-  }
-
-  public void setRelayStateGiven(boolean relayStateGiven) {
-    this.isRelayStateGiven = relayStateGiven;
   }
 }
--- a/plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpRedirectResponse.java
+++ b/plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpRedirectResponse.java
@@ -47,10 +47,8 @@ import static org.apache.cxf.rs.security.saml.sso.SSOConstants.SIG_ALG;
  *
  * </blockquote>
  */
-public class IdpRedirectResponse {
-  private IdpRedirectResponse() {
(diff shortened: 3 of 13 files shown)


plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpRedirectResponse.java
```diff
@@ -128,31 +133,11 @@ public String getSignature() {
     return signature;
   }
```
bakejeyner: Pulled these getters/setters into parent class

test/common/src/main/kotlin/org/codice/compliance/verification/ResponseVerifier.kt
```diff
@@ -15,63 +15,41 @@
...
+        return response.responseDom
     }
 
-    abstract protected fun parseResponse(rawResponse: String): Map<String, String>
-    abstract protected fun decodeResponse(parsedResponse: Map<String, String>): String
-    abstract protected fun getBindingVerifier(responseDom: Node, parsedResponse: Map<String, String>, givenRelayState: Boolean): BindingVerifier
+    abstract protected fun decodeResponse(response: IdpResponse): String
+    abstract protected fun getBindingVerifier(response: IdpResponse): BindingVerifier
 
 }
```
bakejeyner: Don't need the `parseResponse` functions because the parsing is done within the new object when receiving the data from the plugin portion.

plugins/ddf-plugins/src/main/kotlin/org/codice/compliance/saml/plugin/ddf/IdpResponderProvider.kt
```diff
@@ -33,34 +35,17 @@ class IdpResponderProvider : IdpResponder {
...
 
         /*
         * TODO "TODO "Manually change DDF IdP to respond with 302/303 status code for Redirect"
         * When ticket is finished, replace below url building line with:
         * .url(response.getHeader("Location"));
+        *
+        * And delete the html block comment
          */
 
-        /*
```
bakejeyner: Implemented this comment block on line 96

plugins/ddf-plugins/src/main/kotlin/org/codice/compliance/saml/plugin/ddf/IdpResponderProvider.kt
```diff
@@ -104,42 +93,16 @@ class IdpResponderProvider : IdpResponder {
          * ...
          ************************/
 
-        /*
```
bakejeyner: Implemented this comment block on line 97

plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpPostResponse.java
```diff
@@ -70,15 +69,18 @@ public IdpPostResponse build() {
     }
   }
 
-  // General
```
bakejeyner: Pulled these variables and their getters/setters into parent class.

plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpRedirectResponse.java
```diff
@@ -72,19 +69,26 @@ public IdpRedirectResponse build() {
     }
   }
 
-  // General
```
bakejeyner: Pulled these variables and their getters/setters into parent class.

test/idp/src/main/kotlin/org/codice/compliance/web/sso/RedirectLoginTest.kt
```diff
@@ -76,8 +78,11 @@ class RedirectLoginTest : StringSpec() {
                     .`when`()
                     .get(Common.getSingleSignOnLocation(SamlProtocol.REDIRECT_BINDING))
 
+            // Get response from plugin portion
             val idpResponse = getServiceProvider(IdpResponder::class.java).getIdpRedirectResponse(response)
-            verifyResponse(idpResponse, true)
+            idpResponse.isRelayStateGiven = true
```
brjeter: ❗️ You are not setting this in the `PostLoginTest`.
❓  This is nitpicky, but I want this `isRelayStateGiven` property to be `protected` according to my other comments. So my initial thought is to put the set operation in the `IdpResponse` builders and we would pass in the boolean into the `getBINDINGResponse` here in the test. Because only the test knows what the value should be. But then we're relying on the plugin implementers to pass that boolean into the builder which isn't great... We could initialize the Builder object for them and pass it in, but that would be weird IMO. I'm open to ideas, or we may just have to make an exception for making `isRelayStateGiven` `protected`. Thoughts?

plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpResponse.java
```diff
@@ -0,0 +1,52 @@
...
+  public void setDecodedSamlResponse(String decodedSamlResponse) {
+    this.decodedSamlResponse = decodedSamlResponse;
+  }
+
+  public Node getResponseDom() {
+    return responseDom;
+  }
+
+  public void setResponseDom(Node responseDom) {
+    this.responseDom = responseDom;
```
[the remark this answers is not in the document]
bakejeyner: These setters are public on purpose. Some of the values are set by our code after the object is built. For instance: `decodedSamlResponse` is set when we decode the saml response in verification, not during the object build time.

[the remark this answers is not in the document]
brjeter: That's fair. Though that makes me think that we're getting a little confusing here because we're technically storing information for multiple responses at that point, right? I think we should make that fact clear in the object. By naming methods/fields and possibly talking about it in Javadoc.

test/common/src/main/kotlin/org/codice/compliance/verification/binding/PostVerifier.kt
```diff
@@ -13,21 +13,20 @@
...
-class PostVerifier(responseDom: Node, parsedResponse: Map<String, String>, givenRelayState: Boolean): BindingVerifier(responseDom, parsedResponse, givenRelayState) {
+class PostVerifier(val response: IdpPostResponse) : BindingVerifier() {
     /**
      * Verify the response for a post binding
      */
     override fun verifyBinding() {
-        verifySsoPost(responseDom)
-        verifyPostRelayState(parsedResponse[RELAY_STATE], givenRelayState)
+        verifySsoPost(response)
+        verifyPostRelayState(response)
```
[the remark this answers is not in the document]
bakejeyner: This dependency injection would be nice if we ever decided to unit test these methods down the road. @blen-desta , @coyotesqrl , How do you feel?

test/common/src/main/kotlin/org/codice/compliance/verification/binding/RedirectVerifier.kt
```diff
@@ -13,41 +13,40 @@
...
-        verifyRedirectRelayState(parsedResponse[RELAY_STATE], givenRelayState)
-        parsedResponse[SIGNATURE]?.let {
-            verifyRedirectSignature(it, parsedResponse[SAML_RESPONSE], parsedResponse[RELAY_STATE], parsedResponse[SIG_ALG])
-            verifyRedirectDestination(responseDom)
+        verifyRequestParam(response)
+        verifyNoXMLSig(response)
+        verifyRedirectRelayState(response)
+        response.signature?.let {
+            verifyRedirectSignature(response)
+            verifyRedirectDestination(response)
```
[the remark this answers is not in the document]
bakejeyner: See comment above.

test/idp/src/main/kotlin/org/codice/compliance/web/sso/RedirectLoginTest.kt
bakejeyner: Ooo, I can make `isRelayStateGiven` private. I'll do that.

Back to the topic at hand, I chose this way because:
1. I hate flags. I use them so often but I hate them.
2. I believe the `isRelayStateGiven` is information only we know, therefore it should be our responsibility and our's alone to set it
3. Builders _usually_ initialize the builder and build from scratch, therefore I thought passing in a builder for them to use would be kind of odd.

But of course this is open for discussion @blen-desta @coyotesqrl .

library/src/main/kotlin/org/codice/compliance/SAMLComplianceExceptionMessage.kt
```diff
@@ -10,9 +10,11 @@ enum class SAMLComplianceExceptionMessage(val message: String) {
     /***************
      *
      * General
-     *e
```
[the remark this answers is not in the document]
bakejeyner: 🥈

plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpResponse.java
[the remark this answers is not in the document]
bakejeyner: >we're technically storing information for multiple responses at that point, right?
It is the same exact samlResponse that is stored in the built `samlResponse` property, just decoded.

But I always agree with more commentation. I'll try to push something up.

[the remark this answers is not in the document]
brjeter: 👍 Right. It's just various representations of the response. Duh 😄

test/common/src/main/kotlin/org/codice/compliance/verification/binding/PostVerifier.kt
```diff
@@ -13,21 +13,20 @@
...
 
-class PostVerifier(responseDom: Node, parsedResponse: Map<String, String>, givenRelayState: Boolean): BindingVerifier(responseDom, parsedResponse, givenRelayState) {
+class PostVerifier(val response: IdpPostResponse) : BindingVerifier() {
     /**
      * Verify the response for a post binding
      */
     override fun verifyBinding() {
-        verifySsoPost(responseDom)
-        verifyPostRelayState(parsedResponse[RELAY_STATE], givenRelayState)
+        verifySsoPost(response)
```
coyotesqrl: ✏️ No need to pass the `response` down to the functions; they would already have access to the final `response` field on the class. They should also both be private.

test/common/src/main/kotlin/org/codice/compliance/verification/binding/RedirectVerifier.kt
```diff
@@ -13,41 +13,40 @@
...
      * Verify the response for a redirect binding
      */
     override fun verifyBinding() {
-        verifyRequestParam(parsedResponse[SAML_RESPONSE])
-        verifyNoXMLSig(responseDom)
-        verifyRedirectRelayState(parsedResponse[RELAY_STATE], givenRelayState)
-        parsedResponse[SIGNATURE]?.let {
-            verifyRedirectSignature(it, parsedResponse[SAML_RESPONSE], parsedResponse[RELAY_STATE], parsedResponse[SIG_ALG])
-            verifyRedirectDestination(responseDom)
+        verifyRequestParam(response)
```
coyotesqrl: ✏️ Same comment as in `PostVerifier` regarding private methods and not passing the `response` param.

test/common/src/main/kotlin/org/codice/compliance/verification/binding/RedirectVerifier.kt
```diff
@@ -66,7 +65,12 @@ class RedirectVerifier(responseDom: Node, parsedResponse: Map<String, String>, g
...
      * 3.4.4.1 DEFLATE Encoding
      */
-    fun verifyRedirectSignature(signature: String, samlResponse: String?, relayState: String?, sigAlg: String?) {
+    fun verifyRedirectSignature(response: IdpRedirectResponse) {
+        val signature = response.signature
+        val samlResponse = response.samlResponse
+        val relayState = response.relayState
+        val sigAlg = response.sigAlg
+
         val verify: Boolean
```
coyotesqrl: ❗️ Never used.

test/common/src/main/kotlin/org/codice/compliance/verification/binding/RedirectVerifier.kt
```diff
@@ -66,7 +65,12 @@ class RedirectVerifier(responseDom: Node, parsedResponse: Map<String, String>, g
      * Verifies the Signature and SigAlg according to the redirect binding rules in the binding spec
      * 3.4.4.1 DEFLATE Encoding
      */
-    fun verifyRedirectSignature(signature: String, samlResponse: String?, relayState: String?, sigAlg: String?) {
+    fun verifyRedirectSignature(response: IdpRedirectResponse) {
+        val signature = response.signature
```
coyotesqrl: ✏️ Options:
1. Just use the `response.X` values directly in the call below to `SimpleSign().validateSignature`
2. Wrap the call to `SimpleSign().validateSignature` with a `with(response)` block.

I think the latter might actually read a bit clumsy (but not too badly); the former is probably the most clear.

As opposed to the `verifyRedirectRelayState` method, where the local `val`s are used more than once, there really isn't any gain here by saving them off in a new handle.

test/common/src/main/kotlin/org/codice/compliance/verification/binding/RedirectVerifier.kt
```diff
@@ -66,7 +65,12 @@ class RedirectVerifier(responseDom: Node, parsedResponse: Map<String, String>, g
      * Verifies the Signature and SigAlg according to the redirect binding rules in the binding spec
      * 3.4.4.1 DEFLATE Encoding
      */
-    fun verifyRedirectSignature(signature: String, samlResponse: String?, relayState: String?, sigAlg: String?) {
+    fun verifyRedirectSignature(response: IdpRedirectResponse) {
```
coyotesqrl: ✏️ Method can be private.

test/common/src/main/kotlin/org/codice/compliance/verification/binding/RedirectVerifier.kt
```diff
@@ -95,7 +99,9 @@ class RedirectVerifier(responseDom: Node, parsedResponse: Map<String, String>, g
      * 3.4.3 RelayState
      * 3.4.4.1 DEFLATE Encoding
      */
-    fun verifyRedirectRelayState(encodedRelayState: String?, givenRelayState: Boolean) {
+    fun verifyRedirectRelayState(response: IdpRedirectResponse) {
```
coyotesqrl: ✏️ Method can be private.

test/common/src/main/kotlin/org/codice/compliance/verification/binding/RedirectVerifier.kt
```diff
@@ -129,8 +135,8 @@ class RedirectVerifier(responseDom: Node, parsedResponse: Map<String, String>, g
      * Verifies the destination is correct according to the redirect binding rules in the bindinc spec
      * 3.4.5.2 Security Considerations
      */
-    fun verifyRedirectDestination(responseDomElement: Node) {
-        val destination = responseDomElement.attributes.getNamedItem("Destination")?.nodeValue
+    fun verifyRedirectDestination(response: IdpRedirectResponse) {
```
coyotesqrl: ✏️ Method can be private.

test/idp/src/main/kotlin/org/codice/compliance/web/sso/RedirectLoginTest.kt
coyotesqrl: ✏️ Optionally:
```kotlin
            val idpResponse = getServiceProvider(IdpResponder::class.java).getIdpRedirectResponse(response).apply {
                isRelayStateGiven = true
            }
```

test/common/src/main/kotlin/org/codice/compliance/verification/binding/RedirectVerifier.kt
bakejeyner: Good catch.

I chose to go with the first option.
The second option (wrapping with a `when(response)` block nested the `try` loop and looked too clunky. Plus the amount of space it saved with using `this.X` instead of `response.X` wasn't worth it.

test/idp/src/main/kotlin/org/codice/compliance/web/sso/RedirectLoginTest.kt
brjeter: I'd be curious to know what the "best practice" is here for kotlin. Or even what the preference is for experienced kotlin developers. My initial reaction is that the apply is overkill with just one property, but I know that I'm biased towards the Java way of doing things 😄

test/common/src/main/kotlin/org/codice/compliance/verification/binding/PostVerifier.kt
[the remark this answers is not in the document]
coyotesqrl: I'm okay with not testing the test code over much, lest we need to test our test code tests ad infinitum.

test/idp/src/main/kotlin/org/codice/compliance/web/sso/RedirectLoginTest.kt
coyotesqrl: I like that it helps to condition developers into seeing things as more final. Clearly in this case it is a mutable object, but by favoring this idiom we get into a habit of thinking of instances as things we create once and never change.

brjeter: That's a great answer.

plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpResponder.java
```diff
@@ -20,19 +20,38 @@
...
+   *     authn request
+   * @return The response as an {@code IdpRedirectResponse}. The internal builder should be called
+   *     to build the response object:
+   *     <pre>{@code
+   * IdpRedirectResponse.Builder builder = new IdpRedirectResponse.Builder();
+   * builder.httpStatusCode(exampleStatusCode)
+   *     .url(exampleUrl);
+   * return builder.build();
+   * }
+   * where {@code exampleStatusCode} is the http status code returned by the Idp
```
brjeter: ✏️ nit: capital P in `IdP`
(4 different places in this interface)

test/common/src/main/kotlin/org/codice/compliance/utils/TestCommon.kt
```diff
@@ -50,16 +52,23 @@ class TestCommon {
...
+                    .parse(this.decodedSamlResponse.byteInputStream())
                     .documentElement
         }
 
+        /**
+         * Extend {@code Node} to get all nodes under it
+         */
+        fun NodeImpl.getChildren() {
+            return
+        }
```
brjeter: ❗️ This should be removed

test/common/src/main/kotlin/org/codice/compliance/verification/binding/RedirectVerifier.kt
```diff
@@ -82,7 +82,7 @@ class RedirectVerifier(responseDom: Node, parsedResponse: Map<String, String>, g
                 SigErrorCode.INVALID_CERTIFICATE -> throw SAMLComplianceException.create(SAMLBindings_3_1_2_1, message = "The certificate was invalid.", cause = e)
                 SigErrorCode.SIG_ALG_NOT_PROVIDED -> throw SAMLComplianceException.create(SAMLBindings_3_4_4_1_d1, message = "Signature Algorithm not found.", cause = e)
                 SigErrorCode.SIGNATURE_NOT_PROVIDED -> throw SAMLComplianceException.create(SAMLBindings_3_4_4_1_f2, message = "Signature not found.", cause = e)
-                SigErrorCode.INVALID_URI -> throw SAMLComplianceException.create(SAMLBindings_3_4_4_1_d2, message = "The Signature algorithm named $sigAlg is unknown.", cause = e)
+                SigErrorCode.INVALID_URI -> throw SAMLComplianceException.create(SAMLBindings_3_4_4_1_d2, message = "The Signature algorithm named $response.sigAlg is unknown.", cause = e)
```
brjeter: ❗️ Could you manually test this message by forcing it? If IntelliJ is any indication, this will print out the `response` variable value and then the word `.sigAlg`.

bakejeyner: Good catch. Surrounded it in curlys.

plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpResponder.java
```diff
@@ -20,19 +20,38 @@
...
    *
-   * @param originalResponse - the response from the initial POST authn request
-   * @return The response as a string. If a relay state exists concatenate it with a & For example,
-   *     RelayState=%s&SAMLResponse=%s or SAMLResponse=%s&RelayState=%s
+   * @param originalResponse - the original {@code RestAssured} response from the initial POST authn
+   *     request
+   * @return The response as an {@code IdpPostResponse}. The internal builder should be called to
+   *     build the response object:
+   *     <pre>{@code
+   * IdpPostResponse.Builder builder = new IdpPostResponse.Builder();
```
coyotesqrl: ✏️ The normal use of a fluent builder in Java would be more like this:
```java
return new IdpPostResponse.Builder()
    .httpStatusCode(exampleStatusCode)
    .samlForm(exampleSampleForm)
    .build();
```

plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpResponse.java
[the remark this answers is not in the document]
coyotesqrl: ❓ Is there value to holding these other representations at this level, or will they be moved to a decorator at some point?

plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpPostResponse.java
```diff
@@ -70,75 +72,94 @@ public IdpPostResponse build() {
...
+     *
+     * The two key words here are "in" and "hidden"
+     *
+     * Assuming "in" in the above quote means in either the value attribute or in the value
+     * itself.
+     *
+     * And "hidden" means both the SAMLResponse and RelayState MUST be placed in "hidden" form controls
+     */
+    // SAMLResponse portion
+    if (samlResponseNode != null) {
```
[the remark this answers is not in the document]
bakejeyner: Id prefer to keep the null checks, because then I would have to repeatedly call `relayStateNode.getValue` or assign it to a local variable. Those solutions just seem...messier.

Unless, of course, there is a good reason to do so from a best practices perspective.

PR #12, opened 2018-03-16
Switch to gradle

blen-desta:
Updated docs but quick commands:

- `gradle build installDist` - will unzip the created distro under `saml-conformance/distribution/command-line/build/install` and the zip and tar files will be under `saml-conformance/distribution/command-line/build/distributions`
- `gradle build docker` - **will** build the `Dockerfile`.

Next steps:

- Change Jenkins to use gradle instead of maven
- Make the script runnable from any directory (can only run from bin right now)
- Resolve remaining gradle warnings


distribution/command-line/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+applicationDistribution.from('src/main/resources/') {
+    into 'conf'
+}
+
+applicationDistribution.from(project(":plugins:ddf-plugins").file('build/libs')) {
+    into 'plugins'
+}
+
+// Unzip distro must be called specifically (gradle build unzip)
+task unzip(type: Copy) {
```
oconnormi: There actually should already be a task provided by the application plugin called `installDist` that does this :smile:

blen-desta: ☝️ Very true 😆

README.md
```diff
@@ -9,76 +9,92 @@ SAML: https://wiki.oasis-open.org/security/FrontPage
...
-
->NOTE
->
->In order for the test kit to execute properly, you must configure both
-the test kit's and your IdP's/SP's metadata, as well as implement plugins
+To build the project, execute `gradle build` at the project root.
+The `distribution/command-line` module will contain a full package of the deployment after the build.
+
+Tests can be run with the `samlconf` script under `<PATH_TO_PROJECT>/distribution/command-line/build/distributions/samlconf-<VERSION>/bin/`.
+Build the project will create the distribution, to unzip it, execute `gradle build installDist`, which will unzip it and place it under `build/install`.
```
coyotesqrl: ✏️ Since we're already building the unzipped distro, do we really need to provide this direction? Maybe just point out that there are zip and tar archives also available after a full build?

README.md
```diff
@@ -9,76 +9,92 @@ SAML: https://wiki.oasis-open.org/security/FrontPage
...
+       samlconf - Runs the SAML Conformance Tests against an IdP or an SP
+
+SYNOPSIS
+       samlconf [-i path] [--idpMetadata path]
+           [-p path] [--plugins path] 
+
+DESCRIPTION
+       Runs the SAML Conformance Tests which tests the complance of an IdP and/or an SP
+       with the SAML Specifications. If a compliance issue is identified, a 
+       SAMLConformaceException will be throw with an explanation of the error an a direct
```
coyotesqrl: ❗️ s/b "SAMLConformanceException" (missing 'n').
❗️ s/b "...will be thrown..." (missing 'n').
❗️ s/b "...error and a direct..." (missing 'd')

README.md
```diff
@@ -9,76 +9,92 @@ SAML: https://wiki.oasis-open.org/security/FrontPage
...
+       quote from the specification. All of the parameters are optional and if they are 
+       not provided, the default values will use DDF's parameters. All parameters must 
+       be given one times.
```
coyotesqrl: ❗️ s/b "...one time."

README.md
```diff
@@ -9,76 +9,92 @@ SAML: https://wiki.oasis-open.org/security/FrontPage
...
+
+OPTIONS
+       -i path
```
coyotesqrl: ❓ Could we maybe put the alternate options together like they would be in a *nix man page? So instead of two identical descriptions for `-i` and `--idpMetadata` we just get one?

README.md
```diff
@@ -9,76 +9,92 @@ SAML: https://wiki.oasis-open.org/security/FrontPage
...
+           The path to the IdP metadata. If it is not given, the default IdP metadata
+           is /conf/idp-metadata.xml.
+           
+       --idpMetadata path
+           The path to the IdP metadata. If it is not given, the default IdP metadata
+           is /conf/idp-metadata.xml.
+           
+       -p path
+            The path to the pluggable portion. If it is not given, the default plugin
```
coyotesqrl: ✏️ Maybe change to something like "The path to the custom, server-specific plugin implementations. If it is not given, the default plugin directory is..."

README.md
```diff
@@ -9,76 +9,92 @@ SAML: https://wiki.oasis-open.org/security/FrontPage
...
-  `<PATH_TO_PROJECT>/distribution/command-line/src/main/resources/test-sp-metadata.xml`.
+  * Provide your IdP's metadata file path to the `samlconf` script using `-i` or `--idpMetadata`.
+  * Configure your IdP with the test kit's SP metadata from
+  `<PATH_TO_PROJECT>/distribution/command-line/src/main/resources/test-sp-metadata.xml`
+  or `samlconf-1.0-SNAPSHOT/conf/test-sp-metadata.xml` from the distribution.
   
 * **TODO** If testing an SP:
-  * Place your SP's metadata into `sp-metadata.xml` under `<PATH_TO_PROJECT>/distribution/command-line/src/main/resources`.
-  * Copy `test-idp-metadata.xml` under `<PATH_TO_PROJECT>/distribution/command-line/src/main/resources` into your SP.
+  * Provide your SP's metadata file path to the `samlconf` script using `-s` or `--spMetadata`.
```
coyotesqrl: ❓ Is this still valid? This parameter is not included in the man page-like section above.

distribution/command-line/src/main/kotlin/org/codice/ckt/Command.kt
```diff
@@ -2,7 +2,7 @@ package org.codice.ckt
 
 import us.jimschubert.kopper.Parser
 
-const val distroDir = "/distribution/command-line/target/command-line-1.0-SNAPSHOT-bin"
+val samlDist = "${System.getProperty("user.dir")}/.."
```
coyotesqrl: ✏️ Not necessary for this PR, but we should be able to canonicalize this path and simplify.

README.md
```diff
@@ -9,76 +9,92 @@ SAML: https://wiki.oasis-open.org/security/FrontPage
...
+```
+NAME
+       samlconf - Runs the SAML Conformance Tests against an IdP or an SP
+
+SYNOPSIS
+       samlconf [-i path] [--idpMetadata path]
+           [-p path] [--plugins path] 
+
+DESCRIPTION
+       Runs the SAML Conformance Tests which tests the complance of an IdP and/or an SP
```
brjeter: ✏️ "compliance" missing an "i"

README.md
brjeter: Agreed. I'm even confused as to what the point of running `gradle build installDist` is if you've already told me I can run the tests in the sentence above it.

build.gradle
```diff
@@ -0,0 +1,60 @@
...
+    sourceCompatibility = 1.8
+    targetCompatibility = 1.8
+
+    repositories {
+        mavenLocal()
+        mavenCentral()
+    }
+
+    dependencies {
+        compile "org.jetbrains.kotlin:kotlin-reflect"
```
brjeter: ❓ What is this dependency for? Looks like a new one.

distribution/docker/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+
+description = 'Dockerized SAML CKT.'
+
+dependencies {
+    compile project(':distribution:command-line')
+    compile project(':distribution:suites')
+}
+
+// Copy the distribution from the distribution module over to the build module
+task copyDistribuion(type: Copy) {
```
brjeter: ✏️ "copyDistribution" missing a "t"
❓ What is the purpose of doing this copy?

README.md
blen-desta: It doesn't exist yet but that will be what we do. I will add a note that it's not implemented yet on the README.

distribution/command-line/src/main/kotlin/org/codice/ckt/Command.kt
blen-desta: I'll add that to the ticket to fix the script

build.gradle
blen-desta: Yes, that was because of this warning we were getting:

```
w: Runtime JAR files in the classpath should have the same version. These files were found in the classpath:
    /Users/destab/.m2/repository/org/jetbrains/kotlin/kotlin-stdlib-jdk8/1.2.30/kotlin-stdlib-jdk8-1.2.30.jar (version 1.2)
    /Users/destab/.m2/repository/org/jetbrains/kotlin/kotlin-stdlib-jdk7/1.2.30/kotlin-stdlib-jdk7-1.2.30.jar (version 1.2)
    /Users/destab/.m2/repository/org/jetbrains/kotlin/kotlin-reflect/1.1.51/kotlin-reflect-1.1.51.jar (version 1.1)
    /Users/destab/.m2/repository/org/jetbrains/kotlin/kotlin-stdlib/1.2.30/kotlin-stdlib-1.2.30.jar (version 1.2)
w: Consider providing an explicit dependency on kotlin-reflect 1.2 to prevent strange errors
```

distribution/docker/build.gradle
blen-desta: We need to copy the distribution over to the `docker` so the Dockerfile can have access to it. We were doing the same in maven too.

build.gradle
brjeter: Awesome! I was getting tired of seeing those.

PR #17, opened 2018-03-20
Changed ObjectMapper use to Gson

blen-desta:
- Changed ObjectMapper use to Gson
- Added out directories to .gitignore

### DIFF
--- a/plugins/ddf-plugins/src/main/kotlin/org/codice/compliance/saml/plugin/ddf/IdpResponderProvider.kt
+++ b/plugins/ddf-plugins/src/main/kotlin/org/codice/compliance/saml/plugin/ddf/IdpResponderProvider.kt
@@ -13,7 +13,9 @@
  */
 package org.codice.compliance.saml.plugin.ddf
 
-import com.fasterxml.jackson.databind.ObjectMapper
+import com.google.common.reflect.TypeToken
+import com.google.gson.Gson
+import com.google.gson.GsonBuilder
 import com.jayway.restassured.RestAssured
 import com.jayway.restassured.builder.RequestSpecBuilder
 import com.jayway.restassured.response.Response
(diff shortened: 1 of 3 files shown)


plugins/ddf-plugins/src/main/kotlin/org/codice/compliance/saml/plugin/ddf/IdpResponderProvider.kt
```diff
@@ -126,8 +128,9 @@ class IdpResponderProvider : IdpResponder {
                 .replace("window.idpState = ", "")
                 .replace(";", "")
 
-        val queryParams = ObjectMapper()
-                .readValue(idpState, MutableMap::class.java) as MutableMap<String, String>
+        val gson = Gson()
+        GsonBuilder().setPrettyPrinting().create()
```
brjeter: ❓ What does this line do? Seems unnecessary.

blen-desta: It just adds the"pretty print" feature. I can remove it cause we don't really use it right now

plugins/ddf-plugins/src/main/kotlin/org/codice/compliance/saml/plugin/ddf/IdpResponderProvider.kt
```diff
@@ -13,7 +13,9 @@
  */
 package org.codice.compliance.saml.plugin.ddf
 
-import com.fasterxml.jackson.databind.ObjectMapper
+import com.google.common.reflect.TypeToken
+import com.google.gson.Gson
+import com.google.gson.GsonBuilder
```
brjeter: Remove import? Did you run detekt?

PR #26, opened 2018-03-22
ktlint formatter added and resolved findings

idperez:
Adds a formatter to the project and addresses formatting findings. Readme updated as well.

Using the Spotless project (https://github.com/diffplug/spotless/tree/master/plugin-gradle) which has ktlint (https://ktlint.github.io/) support and license header enforcement. 

Running `gradle build` will run the formatter. Any errors can be resolved by running `gradle spotlessApply`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,6 +16,7 @@ buildscript {
 plugins {
     id 'org.jetbrains.kotlin.jvm' version '1.2.30'
     id 'io.gitlab.arturbosch.detekt' version '1.0.0.RC6-4'
+    id 'com.diffplug.gradle.spotless' version '3.10.0'
 }
 
 description = 'SAML Conformance Test Kit'
@@ -41,6 +42,7 @@ subprojects {
     apply plugin: 'kotlin-kapt'
     apply plugin: 'kotlin'
     apply plugin: 'com.bmuschko.docker-remote-api'
+    apply plugin: 'com.diffplug.gradle.spotless'
 
     sourceCompatibility = 1.8
     targetCompatibility = 1.8
@@ -61,6 +63,14 @@ subprojects {
         compile "com.jayway.restassured:rest-assured:$rest_assured_version"
         compile "org.slf4j:slf4j-api:$org_slf4j_version"
         compile "com.google.guava:guava:$guava_version"
+    }
+
+    spotless {
+        kotlin {
+            ktlint()
+
+            licenseHeaderFile rootProject.file('spotless.license.kt')
+        }
     }
 }
 
--- a/test/common/src/main/kotlin/org/codice/compliance/verification/ResponseVerifier.kt
+++ b/test/common/src/main/kotlin/org/codice/compliance/verification/ResponseVerifier.kt
@@ -11,12 +11,10 @@
  * License is distributed along with this program and can be found at
  * <http://www.gnu.org/licenses/lgpl.html>.
  */
-
 package org.codice.compliance.verification
 
 import io.kotlintest.matchers.shouldNotBe
 import org.codice.compliance.SAMLComplianceException
-import org.codice.compliance.SAMLSpecRefMessage.*
 import org.codice.compliance.saml.plugin.IdpPostResponse
 import org.codice.compliance.saml.plugin.IdpRedirectResponse
 import org.codice.compliance.saml.plugin.IdpResponse
@@ -30,6 +28,9 @@ import org.codice.security.sign.Decoder.InflationException.InflErrorCode
 import org.w3c.dom.Node
 import java.io.IOException
 
+/* ktlint-disable no-wildcard-imports */
+import org.codice.compliance.SAMLSpecRefMessage.*
+
 sealed class ResponseVerifier(val response: IdpResponse) {
 
     internal fun verifyResponse(): Node {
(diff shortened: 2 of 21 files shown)


test/common/src/main/kotlin/org/codice/compliance/verification/ResponseVerifier.kt
```diff
@@ -30,6 +28,9 @@ import org.codice.security.sign.Decoder.InflationException.InflErrorCode
 import org.w3c.dom.Node
 import java.io.IOException
 
+/* ktlint-disable no-wildcard-imports */
+import org.codice.compliance.SAMLSpecRefMessage.*
+
```
blen-desta: ❓ Are we not able to import what we need without using wildcard for the `SAMLSpecRefMessage` stuff?

test/idp/src/main/kotlin/org/codice/compliance/web/sso/RedirectLoginTest.kt
```diff
@@ -35,6 +34,9 @@ import org.codice.security.sign.Encoder
 import org.codice.security.sign.SimpleSign
 import java.time.Instant
 
+/* ktlint-disable no-wildcard-imports */
+import org.apache.cxf.rs.security.saml.sso.SSOConstants.*
```
blen-desta: ✏️ We should import the individual `SSOConstants` instead of `suppressing?` this.

idperez: Could the number of SSOConstants used here grow? Importing each one could get out of hand

test/common/src/main/kotlin/org/codice/compliance/verification/ResponseVerifier.kt
idperez: Wildcard imports are not allowed under ktlint so we need a suppressor

coyotesqrl: Wildcards are disallowed in our Java standard; let's be consistent.

https://google.github.io/styleguide/javaguide.html#s3.3-import-statements

test/idp/src/main/kotlin/org/codice/compliance/web/sso/RedirectLoginTest.kt
blen-desta: I don't think it will. As of right now you just need four. `SAML_REQUEST`, `SIG_ALG`, `SIGNATURE` and `RELAY_STATE`

test/common/src/main/kotlin/org/codice/compliance/verification/ResponseVerifier.kt
blen-desta: Yeah, that was a question more for @brjeter on whether or not we can have individual import instead of suppressing it. If doing the individual is too much, I'm okay with the suppression.

test/idp/src/main/kotlin/org/codice/compliance/web/sso/RedirectLoginTest.kt
idperez: Okay, I would say in this case we can import each one. Some of the files it may be overkill

test/common/src/main/kotlin/org/codice/compliance/verification/ResponseVerifier.kt
blen-desta: If it's too many one thing you can do is import `org.codice.compliance.SAMLSpecRefMessage` instead and of using `SAMLCore_3_2_2_a` in the code do `SAMLSpecRefMessage.SAMLCore_3_2_2_a`

coyotesqrl:
Honestly, it's basically invisible to anyone using a modern ide; the
imports block is collapsed.

coyotesqrl:
One last thought on the imports...if we did decide that the imports for consts/enums became too long, we could as @blen-desta suggested, just import `org.codice.compliance.SAMLSpecRefMessage` and then create a [type alias](https://kotlinlang.org/docs/reference/type-aliases.html) to `SAMLSpecRefMessage` to make the references to the enums shorter, something like:
```kotlin
import org.codice.compliance.SAMLSpecRefMessage

typealias SpecMsg = SAMLSpecRefMessage
SpecMsg.SAMLCore_3_2_2_a
```
But again, I don't believe it really matters if the import block grows long.

test/idp/src/main/kotlin/org/codice/compliance/web/sso/RedirectLoginTest.kt
brjeter: I agree with both of you. In this instance, we can import the individual ones but in some edge cases it may be better to suppress.

test/common/src/main/kotlin/org/codice/compliance/verification/ResponseVerifier.kt
brjeter: I'd rather not do that ☝️ I'm fine with the suppression or, to Richard's point, just importing each individual message. Developers rarely look at the import blocks.

build.gradle
```diff
@@ -62,6 +64,14 @@ subprojects {
         compile "org.slf4j:slf4j-api:$org_slf4j_version"
         compile "com.google.guava:guava:$guava_version"
     }
+
+    spotless {
+        kotlin {
+            ktlint()
+
+            licenseHeaderFile rootProject.file('spotless.license.kt')
```
brjeter: ✏️ Shouldn't this be renamed to the "codice license" or something?

PR #28, opened 2018-03-22
Profiles 3.1 - Holder of Key Verification

blen-desta:
```
One or more <ds:KeyInfo> elements MUST be present within the <SubjectConfirmationData> 
element. An xsi:type attribute MAY be present in the <SubjectConfirmationData> element 
and, if present, MUST be set to saml:KeyInfoConfirmationDataType (the namespace prefix 
is arbitrary but must reference the SAML assertion namespace).

Note that in accordance with [XMLSig], each <ds:KeyInfo> element MUST identify a 
single cryptographic key. Multiple keys MAY be identified with separate <ds:KeyInfo> 
elements, such as when different confirmation keys are needed for different relying parties.
```

### DIFF
--- a/test/common/src/main/kotlin/org/codice/compliance/utils/TestCommon.kt
+++ b/test/common/src/main/kotlin/org/codice/compliance/utils/TestCommon.kt
@@ -29,13 +29,15 @@ import org.opensaml.saml.saml2.core.impl.IssuerBuilder
 import org.opensaml.saml.saml2.core.impl.NameIDPolicyBuilder
 import java.io.File
 import java.net.URLClassLoader
-import java.util.*
+import java.util.ServiceLoader
 import kotlin.reflect.KClass
 
 class TestCommon {
     companion object {
         const val XSI = "http://www.w3.org/2001/XMLSchema-instance"
         const val ELEMENT = "http://www.w3.org/2001/04/xmlenc#Element"
+        const val SAML_NAMESPACE = "urn:oasis:names:tc:SAML:2.0:assertion"
+        const val HOLDER_OF_KEY_URI = "urn:oasis:names:tc:SAML:2.0:cm:holder-of-key"
         const val ID = "a1chfeh0234hbifc1jjd3cb40ji0d49"
         const val EXAMPLE_RELAY_STATE = "relay+State"
         const val INCORRECT_RELAY_STATE = "RelayStateLongerThan80CharsIsIncorrectAccordingToTheSamlSpec" +
--- a/test/common/src/main/kotlin/org/codice/compliance/verification/profile/ProfilesVerifier.kt
+++ b/test/common/src/main/kotlin/org/codice/compliance/verification/profile/ProfilesVerifier.kt
@@ -53,7 +53,7 @@ class ProfilesVerifier(private val node: Node) {
                         message = "<ds:KeyInfo> not found within the <SubjectConfirmationData> element.")
 
             keyInfos.forEach {
-                if (it.childNodes.length > 1)
+                if (it.children("KeyValue").size > 1)
                     throw SAMLComplianceException.create(SAMLProfiles_3_1_c,
                             message = "<ds:KeyInfo> must not have multiple values.")
             }
--- a/test/common/src/main/kotlin/org/codice/compliance/verification/profile/ProfilesVerifier.kt
+++ b/test/common/src/main/kotlin/org/codice/compliance/verification/profile/ProfilesVerifier.kt
@@ -1,0 +1,62 @@
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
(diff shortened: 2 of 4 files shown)


test/common/src/main/kotlin/org/codice/compliance/utils/TestCommon.kt
```diff
@@ -29,13 +29,15 @@ import org.opensaml.saml.saml2.core.impl.IssuerBuilder
...
 import java.net.URLClassLoader
-import java.util.*
+import java.util.ServiceLoader
 import kotlin.reflect.KClass
 
 class TestCommon {
     companion object {
         const val XSI = "http://www.w3.org/2001/XMLSchema-instance"
         const val ELEMENT = "http://www.w3.org/2001/04/xmlenc#Element"
+        const val SAML_NAMESPACE = "urn:oasis:names:tc:SAML:2.0:assertion"
```
brjeter: ✏️ Since this has to do with an "assertion", should this be named more specifically `ASSERTION_NAMESPACE`?

test/common/src/main/kotlin/org/codice/compliance/verification/profile/ProfilesVerifier.kt
```diff
@@ -0,0 +1,62 @@
...
+                .filter { it.attributes.getNamedItem("Method").textContent == HOLDER_OF_KEY_URI }
+                .flatMap { it.children("SubjectConfirmationData") }
+
+        subjectConfirmationDataList.forEach {
+            val type = it.attributes.getNamedItemNS(XSI, "type")
+            if (type != null && !type.textContent.contains("KeyInfoConfirmationDataType"))
+                throw SAMLComplianceException.createWithPropertyNotEqualMessage(SAMLProfiles_3_1_b,
+                        "type", type.textContent, "KeyInfoConfirmationDataType")
+
+            if (type.firstChild.namespaceURI != SAML_NAMESPACE)
```
brjeter: ❗️  This seems incorrect to me. What is the `firstChild` of the type attribute? Because we need to basically check what's before the colon in the string literal value. In this example, we need to check what namespace the `saml` prefix is pointing to:
`<SubjectConfirmationData xsi:type="saml:KeyInfoConfirmationDataType">`

test/common/src/main/kotlin/org/codice/compliance/verification/profile/ProfilesVerifier.kt
```diff
@@ -0,0 +1,62 @@
...
+                throw SAMLComplianceException.createWithPropertyNotEqualMessage(SAMLProfiles_3_1_b,
+                        "the namespace prefix", type.firstChild.namespaceURI, SAML_NAMESPACE)
+
+            val keyInfos = it.children("KeyInfo")
+            if (keyInfos.isEmpty())
+                throw SAMLComplianceException.create(SAMLProfiles_3_1_a,
+                        message = "<ds:KeyInfo> not found within the <SubjectConfirmationData> element.")
+
+            keyInfos.forEach {
+                if (it.childNodes.length > 1)
```
brjeter: ❗️ According the the [XMLSig] document that the SAML Spec refers to, a `<KeyInfo>` element could have many child nodes that mean various things. There's a `<KeyValue>` element inside there that actually denotes what the value of the Key is. So you want to find any `<KeyValue>` children and make sure there's just 1.

We're doing this here already: https://github.com/connexta/saml-conformance/blob/c887d29ffd3fc77c504131af5cf0917ae8ee3f78/test/common/src/main/kotlin/org/codice/compliance/verification/core/SamlAssertionsVerifier.kt#L164

test/common/src/main/kotlin/org/codice/compliance/utils/TestCommon.kt
blen-desta: We could. I called it SAML because in core it defines it as 
`saml:` | `urn:oasis:names:tc:SAML:2.0:assertion`

test/common/src/main/kotlin/org/codice/compliance/verification/profile/ProfilesVerifier.kt
blen-desta: I had to debug and check that. The thing is if you get tj=he `value` it returns a string and you have no way of finding the namespace URL. `firstchild` actually returns the value with all the namespace information so we can verify it.

We can' just check what's before the colon because `saml` is a reference to `urn:oasis:names:tc:SAML:2.0:assertion`. DDF can use something else as long as it references the correct URI.

blen-desta: Yeah I agree with you but there was nothing else I could find.

brjeter: Oh yeah. I have no issue with it now, as it is. I'm just commenting on how weird their API is 😄

blen-desta: `tru` `fact`

brjeter: I **do** have issue with that ☝️ comment.

PR #29, opened 2018-03-22
Moved call to CoreVerifier from Tests to Req/Res Verifiers

blen-desta:


blen-desta:
❓ I didn't move it to the Profiles verification I moved it to the Core Request and the Core Response Verification classes. The motivation for that is the fact that both need to run the Core verification and we don't need the test to call Core and Req/Res Verifier. Now the tests just call Req/Res only.

PR #30, opened 2018-03-22
ktlint formatter added and resolved findings

bakejeyner:
Formatter now checks the kotlin code when a build occurs, and formats it with another task.

Also updated Readme and import statements.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,6 +16,7 @@ buildscript {
 plugins {
     id 'org.jetbrains.kotlin.jvm' version '1.2.30'
     id 'io.gitlab.arturbosch.detekt' version '1.0.0.RC6-4'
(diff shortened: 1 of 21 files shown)


build.gradle
```diff
@@ -42,6 +43,7 @@ subprojects {
     apply plugin: 'kotlin-kapt'
     apply plugin: 'kotlin'
     apply plugin: 'com.bmuschko.docker-remote-api'
+    apply plugin: 'com.diffplug.gradle.spotless'
```
blen-desta: ❓ Do you have to define this plugin twice? You're doing it in the plugins block (line 19) and here

bakejeyner: If I take it out I get the error
`Multiple steps with name 'ktlint' for spotless format 'kotlin'`

bakejeyner:
It does now @brjeter !

PR #31, opened 2018-03-22
Added debug logging

blen-desta:
I am using [StaticLog](https://github.com/jupf/staticlog#logging-in-kotlin) which I found on [awsom-kotlin](https://github.com/KotlinBy/awesome-kotlin).

### DIFF
--- a/distribution/command-line/src/main/kotlin/org/codice/ckt/Command.kt
+++ b/distribution/command-line/src/main/kotlin/org/codice/ckt/Command.kt
@@ -13,6 +13,8 @@
  */
 package org.codice.ckt
 
+import de.jupf.staticlog.Log
+import de.jupf.staticlog.core.LogLevel
 import org.codice.compliance.IDP_METADATA_PROPERTY
 import org.codice.compliance.PLUGIN_DIR_PROPERTY
 import org.codice.compliance.TEST_SP_METADATA_PROPERTY
@@ -34,6 +36,10 @@ fun main(args: Array<String>) {
             listOf("plugins"),
             description = "Path to the plugins directory")
 
+    parser.flag("d",
+            listOf("debug"),
+            description = "Turn on debug logs.")
+
     val arguments = parser.parse(args)
 
     val idpMetadata = arguments.option("i")
@@ -45,5 +51,10 @@ fun main(args: Array<String>) {
     System.setProperty(TEST_SP_METADATA_PROPERTY, "$samlDist/conf/test-sp-metadata.xml")
     System.setProperty(PLUGIN_DIR_PROPERTY, pluginDir)
 
+    if (arguments.flag("d")) {
+        Log.logLevel = LogLevel.DEBUG
+    } else {
+        Log.logLevel = LogLevel.INFO
+    }
     org.junit.runner.JUnitCore.main("org.codice.compliance.tests.suites.BasicTestsSuite")
 }
--- a/library/src/main/kotlin/org/codice/compliance/Common.kt
+++ b/library/src/main/kotlin/org/codice/compliance/Common.kt
@@ -13,6 +13,8 @@
  */
 package org.codice.compliance
 
+import de.jupf.staticlog.Log
+import de.jupf.staticlog.core.LogLevel
 import org.codice.security.saml.EntityInformation
 import org.codice.security.saml.IdpMetadata
 import org.codice.security.saml.SPMetadataParser
@@ -126,6 +128,12 @@ fun Node.prettyPrintXml(): String {
     val output = StringWriter()
     transformer.transform(DOMSource(this), StreamResult(output))
     return output.toString()
+}
+
+fun Log.debugWithSupplier(message: () -> String) {
+    if (this.logLevel == LogLevel.DEBUG) {
+        this.debug(message())
+    }
 }
 
 fun String.prettyPrintXml(): String {
--- a/library/src/main/kotlin/org/codice/compliance/Common.kt
+++ b/library/src/main/kotlin/org/codice/compliance/Common.kt
@@ -13,18 +13,21 @@
  */
 package org.codice.compliance
 
-import com.sun.org.apache.xml.internal.serialize.OutputFormat
-import com.sun.org.apache.xml.internal.serialize.XMLSerializer
 import org.codice.security.saml.EntityInformation
 import org.codice.security.saml.IdpMetadata
 import org.codice.security.saml.SPMetadataParser
 import org.codice.security.saml.SamlProtocol
 import org.w3c.dom.Node
-import org.xml.sax.InputSource
 import java.io.File
 import java.io.StringReader
 import java.io.StringWriter
-import javax.xml.parsers.DocumentBuilderFactory
+import java.nio.charset.StandardCharsets
+import javax.xml.transform.OutputKeys
+import javax.xml.transform.Transformer
+import javax.xml.transform.TransformerFactory
+import javax.xml.transform.dom.DOMSource
+import javax.xml.transform.stream.StreamResult
+import javax.xml.transform.stream.StreamSource
 
 const val PLUGIN_DIR_PROPERTY = "saml.plugin.deployDir"
 const val IDP_METADATA_PROPERTY = "idp.metadata"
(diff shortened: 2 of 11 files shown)


test/common/src/main/kotlin/org/codice/compliance/verification/profile/SingleSignOnProfileVerifier.kt
```diff
@@ -30,17 +30,24 @@ import org.codice.compliance.utils.TestCommon.Companion.ID
 import org.codice.compliance.utils.TestCommon.Companion.SP_ISSUER
 import org.codice.compliance.utils.TestCommon.Companion.idpMetadata
 import org.opensaml.saml.saml2.metadata.impl.EntityDescriptorImpl
+import org.slf4j.LoggerFactory
 import org.w3c.dom.Node
 
 class SingleSignOnProfileVerifier(private val response: Node, private val acsUrl: String?) {
+    companion object {
+        private val logger = LoggerFactory.getLogger(SingleSignOnProfileVerifier::class.java)
+    }
```
brjeter: ❓ Is there a reason you setup a logger object here instead of using the `Log.debug` method like in the other places

test/common/src/main/kotlin/org/codice/compliance/verification/binding/PostBindingVerifier.kt
```diff
@@ -36,10 +38,12 @@ class PostBindingVerifier(private val response: IdpPostResponseDecorator) {
      * Verify the response for a post binding
      */
     fun verify() {
+        Log.debug("Starting the Binding Verifications for the POST Binding.")
         decodeAndVerify()
         verifySsoPost()
         verifyPostRelayState()
         verifyPostDestination()
+        Log.debug("Completed the Binding Verifications for the POST Binding.")
```
brjeter: ✏️ I'm not sure I see a lot of value in a ton of these "trace" type logs. With the exception of the tests. I think it's good for us to have that for the tests.

distribution/command-line/src/main/kotlin/org/codice/ckt/Command.kt
```diff
@@ -34,6 +36,10 @@ fun main(args: Array<String>) {
             listOf("plugins"),
             description = "Path to the plugins directory")
 
+    parser.flag("d",
+            listOf("debug"),
+            description = "Turn on debug logs.")
+
```
brjeter: ❓ When running the `runsaml` script, do these logs get printed as well?

library/src/main/kotlin/org/codice/compliance/Common.kt
```diff
@@ -73,6 +79,21 @@ class Common {
...
+            val format = OutputFormat(document).apply {
+                indenting = true
+                indent = 2
+                omitXMLDeclaration = false
+                lineWidth = Integer.MAX_VALUE
+            }
+            val outxml = StringWriter()
+            XMLSerializer(outxml, format).serialize(document)
+            return outxml.toString()
+        }
```
brjeter: 💯

coyotesqrl:
@brjeter I like that idea of using a Builder for the compliance exception. We currently have four factory methods and it's a bit of a mess. I had not expected us to go beyond the _one_ factory method I originally put in place and the legacy of that can be seen in the names of the methods. A builder would clean this up nicely.

library/src/main/kotlin/org/codice/compliance/Common.kt
```diff
@@ -13,12 +13,18 @@
  */
 package org.codice.compliance
 
+import com.sun.org.apache.xml.internal.serialize.OutputFormat
```
coyotesqrl: ✏️ Importing `com.sun.*` packages is generally frowned upon, as there are no guarantees what JVM will be run by an end user. Is there a reason these are being used and not some external, more general library?

test/common/src/main/kotlin/org/codice/compliance/verification/binding/PostBindingVerifier.kt
coyotesqrl: Agreed.

library/src/main/kotlin/org/codice/compliance/Common.kt
```diff
@@ -73,6 +79,21 @@ class Common {
                     ?.first { it.binding == binding }
                     ?.location
         }
+
+        fun prettyPrintXml(xml: String): String {
```
coyotesqrl: ✏️ This is a pretty specific function (with a nicely specific/clear name) to be performed on a String; it is a good candidate for an extension function.

test/common/src/main/kotlin/org/codice/compliance/verification/profile/SingleSignOnProfileVerifier.kt
coyotesqrl: ❓ How do those other logging messages print out? This one, I assume, clearly indicates it's using the `SingleSignOnProfileVerifier` logger; do the others have some generic logger source? Do we care?

test/idp/src/main/kotlin/org/codice/compliance/web/sso/PostLoginTest.kt
```diff
@@ -64,14 +66,17 @@ class PostLoginTest : StringSpec() {
                 }
             }.apply { SimpleSign().signSamlObject(this) }
 
-            return authnRequestToString(authnRequest)
+            val authnRequestString = authnRequestToString(authnRequest)
+            Log.debug(prettyPrintXml(authnRequestString))
```
coyotesqrl: ❓ Is there either a conditional test - `isDebugEnabled`- or a version of the methods that takes a lambda? Otherwise, we're going to perform the pretty print formatting every time, even when the log level isn't set to debug and we don't print the log message.

test/common/src/main/kotlin/org/codice/compliance/verification/profile/SingleSignOnProfileVerifier.kt
blen-desta: 😞 My bad that's leftover code from my many attempts at using other debuggers.

test/common/src/main/kotlin/org/codice/compliance/verification/binding/PostBindingVerifier.kt
blen-desta: Do you mean keep just the starting post/redirect test logs or move these to the tests?

distribution/command-line/src/main/kotlin/org/codice/ckt/Command.kt
blen-desta: If you provide the `-d` flag, yes

library/src/main/kotlin/org/codice/compliance/Common.kt
blen-desta: I will look for alternatives

test/common/src/main/kotlin/org/codice/compliance/verification/binding/PostBindingVerifier.kt
brjeter: No I mean get rid of this line and the similar ones but keep the ones you already have in the tests.

test/common/src/main/kotlin/org/codice/compliance/verification/binding/RedirectBindingVerifier.kt
```diff
@@ -63,6 +66,7 @@ class RedirectBindingVerifier(private val response: IdpRedirectResponseDecorator
             verifyRedirectSignature()
             verifyRedirectDestination()
         }
+        Log.debug("Starting the Binding Verifications for the Redirect Binding.")
```
bakejeyner: Should be "**Ending** the Binding Verifications for the Redirect Binding.".

test/idp/src/main/kotlin/org/codice/compliance/web/sso/PostLoginTest.kt
blen-desta: There isn't an `isDebugEnabled` type of method. I can get the log level and check if it's `DEBUG`. 
Looking at other kotlin-based loggers none of them have a log enabled method. Not sure if it's the same with this logger but another one I saw explicitly mentions that you do not need to check if a log level is enabled because it does that check internally.

test/common/src/main/kotlin/org/codice/compliance/verification/profile/SingleSignOnProfileVerifier.kt
blen-desta: It just has one static logger. 

> It is for formatting standard output comfortably without the need to construct a Logger object. But it's also no problem to create one.

PR #34, opened 2018-03-26
Cleans up SAMLComplianceException factory methods

coyotesqrl:
Also adds an optional Node to be pretty-printed in the case of an exception being thrown.

### DIFF
--- a/library/src/main/kotlin/org/codice/compliance/SAMLComplianceException.kt
+++ b/library/src/main/kotlin/org/codice/compliance/SAMLComplianceException.kt
@@ -13,41 +13,52 @@
  */
 package org.codice.compliance
 
+import org.w3c.dom.Node
+
 class SAMLComplianceException : Exception {
 
     private constructor(message: String) : super(message)
     private constructor(message: String, cause: Throwable) : super(message, cause)
 
     companion object {
-        fun create(vararg codes: SAMLSpecRefMessage, message: String, cause: Throwable? = null):
+        fun create(vararg codes: SAMLSpecRefMessage,
+                   message: String,
+                   cause: Throwable? = null,
+                   node: Node? = null):
                 SAMLComplianceException {
             val samlExceptions = codes.map(Companion::readCode)
                     .fold("SAML Specification References:\n") { acc, s ->
                         "$acc\n$s"
                     }
 
-            if (cause != null) {
-                return SAMLComplianceException("$message\n\n$samlExceptions\n", cause)
+            return if (cause != null) {
+                SAMLComplianceException("$message\n\n$samlExceptions\n\n${node?.prettyPrintXml() ?: ""}\n", cause)
             } else {
-                return SAMLComplianceException("$message\n\n$samlExceptions\n")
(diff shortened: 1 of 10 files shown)


library/src/main/kotlin/org/codice/compliance/SAMLComplianceException.kt
```diff
@@ -13,41 +13,52 @@
...
+
 class SAMLComplianceException : Exception {
 
     private constructor(message: String) : super(message)
     private constructor(message: String, cause: Throwable) : super(message, cause)
 
     companion object {
-        fun create(vararg codes: SAMLSpecRefMessage, message: String, cause: Throwable? = null):
+        fun create(vararg codes: SAMLSpecRefMessage,
+                   message: String,
```
bakejeyner: ❓ message is now required in `create`?

library/src/main/kotlin/org/codice/compliance/SAMLComplianceException.kt
```diff
@@ -13,41 +13,52 @@
...
         }
 
-        fun createWithPropertyReqMessage(section: String, property: String, parent: String): SAMLComplianceException {
-            return SAMLComplianceException("$section: $property is required in $parent.")
-        }
-
-        fun createWithPropertyInvalidMessage(code: SAMLSpecRefMessage, property: String, propertyValue: String?):
-                SAMLComplianceException {
-            return SAMLComplianceException("The $property value of $propertyValue is invalid.\n\n${readCode(code)}")
+        fun createWithPropertyReqMessage(section: String,
```
bakejeyner: ❓ Should we specify that `property` in this context means a property in an xml, and not some other property like in an object?

Maybe something like:
```
createWithXmlPropertyReqMessage
```

library/src/main/kotlin/org/codice/compliance/SAMLComplianceException.kt
```diff
@@ -13,41 +13,52 @@
...
-                                              property: String,
-                                              propertyValue: String?,
-                                              otherValue: String?): SAMLComplianceException {
-            return SAMLComplianceException("The $property value of $propertyValue is not equal to $otherValue." +
-                    "\n\n${readCode(code)}")
+        @Suppress("LongParameterList")
+        fun createWithPropertyMessage(code: SAMLSpecRefMessage,
+                                      property: String,
+                                      propertyValue: String?,
+                                      otherValue: String? = null,
```
blen-desta: ✏️ Can we rename `propertyValue` and `otherValue` it's not intuitive. Something like `actual` and `expected`.

library/src/main/kotlin/org/codice/compliance/SAMLComplianceException.kt
coyotesqrl: It always was. The only changes here are to add the `node` and to reformat the method to put each param on a line of its own.

library/src/main/kotlin/org/codice/compliance/SAMLComplianceException.kt
coyotesqrl: I can make that change.

PR #35, opened 2018-03-26
Google java formatter added

idperez:
Google java formatter added (https://github.com/sherter/google-java-format-gradle-plugin), readme updated, formatted all .java files. 

Run `gradle goJF` to format 
Run `gradle verGJF` to verify formatting


README.md
```diff
@@ -51,7 +51,7 @@ for the user-handled portions of SAML profiles. See "Metadata" and "Plugins" for
 ### Formatting
 If during development the build fails due to `format violations` run the following command to format:
 
-    gradle spotlessApply
+    gradle spotlessApply goJF
```
bakejeyner: ✏️  This is a super nit-pick, but is there a way to change these command line arguments? I just feel like `goJF` seems non-intuitive. I could be alone in that opinion though.

coyotesqrl:
Needs rebase.

README.md
idperez: The spotless verify isn't mentioned either for kotlin. Most people would rather resolve style errors rather than see what errors there are. Don't want to over complicate the readme. I agree the goJF isn't intuitive at all. I would like to add a `format` task that would run both spotlessApply and goJF eventually.

PR #39, opened 2018-03-26
Switch to using tripple quotes in SAMLSpecRefMessage.kt

bakejeyner:
Switch to using raw kotlin triple quote strings for our SAML Specification Reference Messages.


build.gradle
```diff
@@ -96,4 +96,4 @@ detekt {
         config = "$projectDir/detekt.yml"
         filters = ".*/resources/.*,.*/tmp/.*"
     }
-}
\ No newline at end of file
+}
```
bakejeyner: Added a newline after this brace because it's our standard. Also because I creep.

library/src/main/kotlin/org/codice/compliance/SAMLSpecRefMessage.kt
```diff
@@ -19,378 +19,378 @@ package org.codice.compliance
  *  - The usage of enums is less error-prone than String literals
  *  - There is no need to have this list be as dynamic as a properties file at runtime
  */
-@Suppress("StringLiteralDuplication")
+@Suppress("""StringLiteralDuplication""")
```
brjeter: Derp

library/src/main/kotlin/org/codice/compliance/SAMLSpecRefMessage.kt
```diff
@@ -19,378 +19,378 @@ package org.codice.compliance
...
-            "- urn:oasis:names:tc:SAML:2.0:status:Requester \n " +
-            "- urn:oasis:names:tc:SAML:2.0:status:Responder \n " +
-            "- urn:oasis:names:tc:SAML:2.0:status:VersionMismatch"),
+    SAMLCore_3_2_2_2("""The value of the topmost <StatusCode> element MUST be from the top-level list provided in this """ +
+            """section.""" +
+            "\n" + """The permissible top-level <StatusCode> values are as follows:""" +
+            "\n" + """- urn:oasis:names:tc:SAML:2.0:status:Success""" +
+            "\n" + """- urn:oasis:names:tc:SAML:2.0:status:Requester""" +
+            "\n" + """- urn:oasis:names:tc:SAML:2.0:status:Responder""" +
+            "\n" + """- urn:oasis:names:tc:SAML:2.0:status:VersionMismatch"""),
```
brjeter: ✏️ This is nitpicky, but it seems like we could make an exception for cases like this where we don't need to escape any characters but we do need to be able to put in a newline.

build.gradle
coyotesqrl: This is not creep. Missing this newline would/should cause a build failure due to static analysis.

library/src/main/kotlin/org/codice/compliance/SAMLSpecRefMessage.kt
coyotesqrl: Agreed. Let's not be overzealous.

library/src/main/kotlin/org/codice/compliance/SAMLSpecRefMessage.kt
coyotesqrl: Agreed. I'd suggested this yesterday and now that I'm seeing it I immediately regret that. 😄

library/src/main/kotlin/org/codice/compliance/SAMLSpecRefMessage.kt
```diff
@@ -205,197 +205,197 @@ enum class SAMLSpecRefMessage(val message: String) {
...
 
     /***************
      *
      * XML Signature Syntax and Processing
      *
      ***************/
 
-    XMLSignature_4_5("compliant versions [of KeyInfo] MUST implement KeyValue")
+    XMLSignature_4_5("""compliant versions [of KeyInfo] MUST implement KeyValue""")
 }
```
blen-desta: ❓ Do you not need a new line at the end?

bakejeyner: I do. It's odd on github.

PR #40, opened 2018-03-26
Refactored duplicate code

blen-desta:
IntelliJ was flagging this as duplicate code

### DIFF
--- a/plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpPostResponse.java
+++ b/plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpPostResponse.java
@@ -133,22 +133,19 @@ public class IdpPostResponse extends IdpResponse {
      *
      * And "hidden" means both the SAMLResponse and RelayState MUST be placed in "hidden" form controls
      */
-    // SAMLResponse portion
-    if (samlResponseForm != null) {
-      if (isNotEmpty(samlResponseForm.value())) {
-        samlResponse = samlResponseForm.value();
-      } else if (isNotEmpty(samlResponseForm.attributes().get(VALUE))) {
-        samlResponse = samlResponseForm.attributes().get(VALUE);
-      }
-    }
 
-    // RelayState portion
-    if (relayStateForm != null) {
-      if (isNotEmpty(relayStateForm.value())) {
-        relayState = relayStateForm.value();
-      } else if (isNotEmpty(relayStateForm.attributes().get(VALUE))) {
-        relayState = relayStateForm.attributes().get(VALUE);
+    samlResponse = extractValue(samlResponseNode);
+    relayState = extractValue(relayStateNode);
+  }
+
+  private String extractValue(Node node) {
+    if (node != null) {
+      if (isNotEmpty(node.value())) {
+        return node.value();
+      } else if (isNotEmpty(node.attributes().get(VALUE))) {
+        return node.attributes().get(VALUE);
       }
     }
+    return null;
   }
 }
--- a/plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpPostResponse.java
+++ b/plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpPostResponse.java
@@ -134,8 +134,8 @@ public class IdpPostResponse extends IdpResponse {
      * And "hidden" means both the SAMLResponse and RelayState MUST be placed in "hidden" form controls
      */
 
-    samlResponse = extractValue(samlResponseNode);
-    relayState = extractValue(relayStateNode);
+    samlResponse = extractValue(samlResponseForm);
+    relayState = extractValue(relayStateForm);
   }
 
   private String extractValue(Node node) {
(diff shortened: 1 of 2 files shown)


plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpPostResponse.java
```diff
@@ -128,32 +128,28 @@ private void parseAndSetFormValues(Node samlResponseForm) {
...
+      if (isNotEmpty(node.value())) {
+        return node.value();
+      } else if (isNotEmpty(node.attributes().get(VALUE))) {
+        return node.attributes().get(VALUE);
       }
     }
+    return null;
+  }
+
+  private boolean extractType(Node node) {
```
bakejeyner: ✏️ I feel like this name doesn't quite describe its function. Maybe `isTypeHidden()` instead?

Also might be cleaner written as:
```
private boolean isTypeHidden(Node node) {
    return node !=null
        && "hidden".equals(node.getAttribute(TYPE))
}
```

plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpPostResponse.java
```diff
@@ -133,22 +133,19 @@ private void parseAndSetFormValues(Node responseForm) {
      *
```
blen-desta: @bakejeyner do we still need this comment ☝️ since we're not dealing with hidden fields anymore?

bakejeyner: Correct.

plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpPostResponse.java
```diff
@@ -133,22 +133,19 @@ private void parseAndSetFormValues(Node responseForm) {
...
-        relayState = relayStateForm.attributes().get(VALUE);
+    samlResponse = extractValue(samlResponseForm);
+    relayState = extractValue(relayStateForm);
+  }
+
+  private String extractValue(Node node) {
+    if (node != null) {
+      if (isNotEmpty(node.value())) {
+        return node.value();
+      } else if (isNotEmpty(node.attributes().get(VALUE))) {
```
bakejeyner: I forgot to add a null check on that `else if` 😀. A NPE could be thrown if `attributes()` is null. A possible refactor could be:
```
private String extractValue(Node node) {
  if (node != null) {
    if (isNotEmpty(node.value())) {
      return node.value();
    } else if (node.attributes() != null && isNotEmpty(node.attributes().get(VALUE))) {
      return node.attributes().get(VALUE);
    }
    return null;
}
```

plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpPostResponse.java
bakejeyner: Good catch! We do not need that comment anymore

plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpPostResponse.java
brjeter: Couldn't you just do this?
```
private String extractValue(Node node) {
  if (node != null) {
    if (isNotEmpty(node.value())) {
      return node.value();
    } else {
      return node.attributes()?.get(VALUE);
    }
}
```
The else will just return null here if `attributes()` is null and will return null if there are no values associated with the `VALUE` key in `attributes().get(VALUE)`.

brjeter: Just kidding you can't do that because this is Java, but just for fun you could even simplify it further by using the Elvis operator (if this was Kotlin)
```
private String extractValue(Node node) {
  return node?.value() ?: node.attributes()?.get(VALUE)
}
```

plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpPostResponse.java
```diff
@@ -133,22 +133,19 @@ private void parseAndSetFormValues(Node responseForm) {
...
-      if (isNotEmpty(relayStateForm.value())) {
-        relayState = relayStateForm.value();
-      } else if (isNotEmpty(relayStateForm.attributes().get(VALUE))) {
-        relayState = relayStateForm.attributes().get(VALUE);
+    samlResponse = extractValue(samlResponseForm);
+    relayState = extractValue(relayStateForm);
+  }
+
+  private String extractValue(Node node) {
+    if (node != null) {
```
coyotesqrl: ✏️ Invert to simplify and reduce complexity.

plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpPostResponse.java
coyotesqrl: Since this is Java, sadly, I would suggest this:
```java
  private String extractValue(Node node) {
    if (node == null) {
      return null;
    }

    if (isNotEmpty(node.value())) {
      return node.value();
    }

    if (node.attributes() == null) {
      return null;
    }
    
    if (isNotEmpty(node.attributes().get(VALUE))) {
      return node.attributes().get(VALUE);
    }
    
    return null;
  }
```

plugins/plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpPostResponse.java
blen-desta: I removed that cause it didn't exist when I rebased

PR #41, opened 2018-03-27
Fix the waiting issue with docker ddf

blen-desta:
- Fix the waiting issue with docker ddf
    * querying for `profile`minimum` isn't working so switching to checking the availability of `security-idp`
    * adding a check for the availability of `/services/idp/login/metadata`
- Changed script to use the gradle generated script
- Updated metadata config


distribution/docker/wait.sh
```diff
@@ -24,7 +26,7 @@ while [ ${cmd} -ne 1 ]
 do
     >&2 echo "DDF is NOT up - sleeping for 10 seconds"
     sleep 10s
-    cmd=$(curl -s -k -i -X GET "https://${_sut_host}:${_sut_port}/admin/jolokia/exec/org.apache.karaf:type=feature,name=root/infoFeature(java.lang.String)/profile-${_target_profile}" | grep -i '"Installed":true' | wc -l)
+    cmd=$(curl -s -k -i -X GET "https://${_sut_host}:${_sut_port}/admin/jolokia/exec/org.apache.karaf:type=feature,name=root/infoFeature(java.lang.String)/${_target_feature}" | grep -i '"Installed":true' | wc -l)
```
coyotesqrl: ❓ Since the commands outside and inside the do loop are identical, could we save them off to be exec'd instead of having to maintain two lines in the script?

blen-desta: I tried to do that last time and that wasn't working very well. I can try again

blen-desta:
Waiting for CI build to merge.

distribution/docker/etc/org.codice.ddf.security.idp.server.IdpEndpoint.config
```diff
@@ -2,6 +2,6 @@ expirationTime=I"30"
...
...S66M\=\ \ \ \ \ \ \ \ \ \ \ </ds:X509Certificate>\ \ \ \ \ \ \ \ \ </ds:X509Data>\ \ \ \ \ \ \ </ds:KeyInfo>\ \ \ \ \ </md:KeyDescriptor>\ \ \ \ \ <md:SingleLogoutService\ Binding\=\"urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect\"\ Location\=\"https://samlhost:8993/services/saml/logout/redirect\"/>\ \ \ \ \ <md:SingleLogoutService\ Binding\=\"urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST\"\ Location\=\"https://samlhost:8993/services/saml/logout/post\"/>\ \ \ \ \ <md:SingleLogoutService\ Binding\=\"urn:oasis:names:tc:SAML:2.0:bindings:SOAP\"\ Location\=\"https://samlhost:8993/services/saml/logout/soap\"/>\ \ \ \ \ <md:AssertionConsumerService\ Binding\=\"urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect\"\ Location\=\"https://samlhost:8993/services/saml/sso/redirect\"\ index\=\"0\"/>\ \ \ \ \ <md:AssertionConsumerService\ Binding\=\"urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST\"\ Location\=\"https://samlhost:8993/services/saml/sso/post\"\ index\=\"1\"/>\ \ \ \ \ <md:AssertionConsumerService\ Binding\=\"urn:oasis:names:tc:SAML:2.0:bindings:PAOS\"\ Location\=\"https://samlhost:8993/services/saml/sso/paos\"\ index\=\"2\"/>\ \ \ </md:SPSSODescriptor>\ </md:EntityDescriptor>", \
```
coyotesqrl: ❓ Does this need to be on one line? I know it doesn't really matter but it's _really_ long.

blen-desta: I just copied it directly from the etc file. I didn't want to tamper with these config files. 

Is there no problem if I added a new line? Do I have to escape something?

PR #44, opened 2018-03-27
Removed samlconf from filename of metadata file

blen-desta:


distribution/command-line/src/main/kotlin/org/codice/ckt/Command.kt
```diff
@@ -43,7 +43,7 @@ fun main(args: Array<String>) {
     val arguments = parser.parse(args)
 
     val idpMetadata = arguments.option("i")
-            ?: "$samlDist/conf/samlconf-idp-metadata.xml"
+            ?: "$samlDist/conf/idp-metadata.xml"
```
brjeter: ❓ Was this just not working before if you didn't provide an explicit argument for the idp metadata?
❓ Could someone explain how this xml file gets renamed from `samlconf-idp-metadata.xml` to `idp-metadata.xml` when we do a gradle build?

coyotesqrl: It doesn't get renamed. It was moved to that new filename yesterday in #38 and this change was overlooked.

blen-desta: ☝️

PR #45, opened 2018-03-28
Verification for the error status codes

blen-desta:
For @bakejeyner to work with. Must be tested with DDF before merging.


test/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
```diff
@@ -15,17 +15,47 @@ package org.codice.compliance.verification.core
...
+            if (status == TestCommon.SUCCESS
+                    && !TestCommon.TOP_LEVEL_STATUS_CODES.contains(status)
+                    && !TestCommon.SECOND_LEVEL_STATUS_CODES.contains(status)) {
+                throw SAMLComplianceException.createWithPropertyMessage(samlErrorCode,
+                        property = "Status Code",
+                        actual = statusCodes[0].attributes.getNamedItem("Value").textContent,
+                        expected = TestCommon.TOP_LEVEL_STATUS_CODES.toString()
+                                .replace("urn:oasis:names:tc:SAML:2.0:status:Success", "")
+                                + TestCommon.SECOND_LEVEL_STATUS_CODES.toString())
+            }
```
brjeter: ❗️ This is technically not verifying the Spec exactly. I don't think we should be looking at second level status codes at all because not only are they optional but the spec says `System entities are free to define more specific status codes by defining appropriate URI references.`. i.e. those second level status codes could be anything. I'd propose these changes:
- Change line 37 to only call `node.children()` instead of `node.allChildren()`.
- Check that there's only one statusCode element that comes out of it.
- Make sure that statusCode is equal to one of the top-level error codes

brjeter: Separate note, but I'm not sure if it's appropriate to change in this PR:

I found this little tidbit about status codes which makes me think we may want to be able to verify the response has a specific status code. In scenarios where we give the IdP bad data, it sounds like the IdP must respond a certain way. From Core 3.2.1:
```
If a SAML responder deems a request to be invalid according to SAML syntax or processing rules,
then if it responds, it MUST return a SAML response message with a <StatusCode> element
with the value urn:oasis:names:tc:SAML:2.0:status:Requester. In some cases, for example
during a suspected denial-of service attack, not responding at all may be warranted.
```

blen-desta: You can have multiple status codes. 
```
If the identity provider cannot or will not satisfy the request, it MUST respond 
with a <Response>message containing an appropriate error status code or 
codes.
```
I opted for doing `node.allChildren()` for that reason. I can verify that there is one `Status` element and do `node.children("Status")[0].children("Status Code)` instead.

I don't see where it says the top level status codes are mandatory and the second level codes are optional. Granted they can specify their own and we can't really check against that. I just didn't want to make the assumption that they have to specify a top-level status code.

blen-desta: This is the case where we provide it bad data. Should I just check that `urn:oasis:names:tc:SAML:2.0:status:Requester` exists?

brjeter: In Core it explains all the rules about Status Codes. Mainly section `3.2.2.2 
Element <StatusCode>`. The part that specifies that a top-level code is required:
```
Value [Required]
The status code value. This attribute contains a URI reference.
The value of the topmost <StatusCode> element MUST be
from the top-level list provided in this section.
```

As for whether or not they could have multiple top-level status codes, I'm not entirely sure based on the Core spec ☹️

brjeter: Yeah maybe we have convenience methods like `verifyRequesterError` and `verifyVersionMismatch` for the two errors that we can actually expect in various scenarios. I don't think we would need a `verifyResponderError` until we do SP tests.

test/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
```diff
@@ -20,12 +20,59 @@ import org.codice.compliance.SAMLCore_5_4_2_a
...
+     * This should be called explicitly if an error of this type is expected.
+     */
+    fun verifyVersionMismatchError(samlErrorCode: SAMLSpecRefMessage) {
+        verifyAndReturnStatus(samlErrorCode).children("Status Code").forEach {
+            val code = it.attributes.getNamedItem("Value").textContent
+            if (code != VERSION_MISMATCH)
+                throw SAMLComplianceException.createWithPropertyMessage(samlErrorCode,
+                        property = "Status Code",
+                        actual = code,
+                        expected = REQUESTER,
```
brjeter: ❗️ This should say `VERSION_MISMATCH` instead of `REQUESTER`. Also, couldn't you put all of this logic into the common private method below and pass in the URI you're expecting? i.e. `VERSION_MISMATCH` or `REQUESTER`.

test/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
```diff
@@ -20,12 +20,59 @@ import org.codice.compliance.SAMLCore_5_4_2_a
...
+
+    /**
+     * Verifies that a response has a status code of urn:oasis:names:tc:SAML:2.0:status:Requester.
+     * This should be called explicitly if an error of this type is expected.
+     */
+    fun verifyRequesterError(samlErrorCode: SAMLSpecRefMessage) {
+        verifyAndReturnStatus(samlErrorCode).children("Status Code").forEach {
+            val code = it.attributes.getNamedItem("Value").textContent
+            if (code != REQUESTER)
+                throw SAMLComplianceException.createWithPropertyMessage(samlErrorCode,
```
bakejeyner: ✏️ I think it would be a good idea to put in some extra SAMLSpecRefMessages here. The ones that say things like
```
HTTP interactions during the message exchange MUST NOT use HTTP error status codes to indicate
failures in SAML processing, since the user agent is not a full party to the SAML protocol exchange.
```
So that if the IdP errors, the person running the test knows what part of the spec we are testing, along with the fact that they have to return a saml error response.

test/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
```diff
@@ -20,12 +20,59 @@ import org.codice.compliance.SAMLCore_5_4_2_a
...
+                        property = "Status Code",
+                        actual = code,
+                        expected = REQUESTER,
+                        node = node)
+        }
+    }
+
+    private fun verifyAndReturnStatus(samlErrorCode: SAMLSpecRefMessage): Node {
+        val status = node.children("Status")
+        if (status.size != 1)
```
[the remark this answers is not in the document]
blen-desta: No there should be one status that has multiple status codes. I think 🤔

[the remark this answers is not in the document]
bakejeyner: Ya I dunno either. I read through that thing like 3 times and still didn't understand that element of it. I think either way is fine!

test/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
blen-desta: So my one thing with that is, it's not part of the Core spec and by adding that we go away from the whole "separate verification by document" thing.

test/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
blen-desta: Oops

test/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
bakejeyner: Good point. Plus we SHOULD (<- now I'm sounding like the spec) be calling the http status code check beforehand anyway.

bakejeyner: Ooo, this is part of the core document and may be beneficial to put in the error codes:
```
If a SAML responder deems a request to be invalid according to SAML syntax or processing rules, then if it responds, it MUST return a SAML response message with a <StatusCode> element with the value urn:oasis:names:tc:SAML:2.0:status:Requester.
```

blen-desta: True but now that it's generalized, you have to add that to the SAML Messages and pass it in if your test is expecting the `Requester` status code

bakejeyner: Or we can be sneaky...
`
If a SAML responder deems a request to be invalid according to SAML syntax or processing rules, then if it responds, it MUST return a SAML response message with a <StatusCode> element
`
or something of the sort.

bakejeyner: I just feel like we should them:
- What part of the spec we were testing when this happened
- And what they should do and why they should do it, according to the spec.

brjeter: @blen-desta and @bakejeyner what if, only in the Requester error message, we don't take a SamlSpecRefMessage, but instead just statically use the paragraph you referenced above ☝️ ?

brjeter: I care less about the version mismatch because that should only really affect like one test...

bakejeyner: I feel like it's a good idea to tell them where in the spec we were negative testing when that happened, as well as that above message. But that's just my opinion.

blen-desta: That is what the code does. I prints the above message and whatever message you provide it (which will be "where in the spec we were negative testing")

test/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
```diff
@@ -15,17 +15,51 @@ package org.codice.compliance.verification.core
...
+                if (expectedStatusCode == REQUESTER)
+                    exceptions = arrayOf(samlErrorCode, SAMLCore_3_2_1_d)
+                throw SAMLComplianceException.createWithPropertyMessage(*exceptions,
+                        property = "Status Code",
+                        actual = code,
+                        expected = expectedStatusCode,
+                        node = status[0])
+            }
+        }
+    }
```
brjeter: I'm not going to lie, this was not what I was thinking. Was thinking we'd have a separate convenience method for the requester with it's own special error message logic. I think I like this better 😄

library/src/main/kotlin/org/codice/compliance/SAMLComplianceException.kt
```diff
@@ -49,17 +49,21 @@ class SAMLComplianceException : Exception {
...
-        fun createWithPropertyMessage(code: SAMLSpecRefMessage,
+        fun createWithPropertyMessage(vararg codes: SAMLSpecRefMessage,
                                       property: String,
                                       actual: String?,
                                       expected: String? = null,
                                       node: Node? = null): SAMLComplianceException {
+            val samlExceptions = codes.map(Companion::readCode)
+                    .fold("SAML Specification References:\n") { acc, s ->
+                        "$acc\n$s"
+                    }
```
brjeter: Skipping over this because I'm in a rush and it takes me a second to wrap my head around the fold thing 😉

test/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
```diff
@@ -15,17 +15,51 @@ package org.codice.compliance.verification.core
...
+     * This should be called explicitly if an error is expected.
+     *
+     * @param samlErrorCode - The error code you wish to use for the SAMLComplianceException
+     * @param expectedStatusCode - the uri of the expected status code.
+     * For example, urn:oasis:names:tc:SAML:2.0:status:Requester.
+     */
+    fun verifyErrorStatusCode(samlErrorCode: SAMLSpecRefMessage, expectedStatusCode: String) {
+        val status = node.children("Status")
+        if (status.size != 1)
+            throw SAMLComplianceException.createWithPropertyMessage(samlErrorCode,
```
bakejeyner: ✏️ Could make sense to use `*exceptions` here as well.

library/src/main/kotlin/org/codice/compliance/SAMLComplianceException.kt
bakejeyner: This was a good resource http://kotlination.com/kotlin/kotlin-fold-example-list-map.

From my understanding, the `.fold` method is like a `forEach` that keeps appending/adding to a single value.
So in this case, `"SAML Specification References:\n"` is the initial value that is stored in the lambda variable `acc`. The other lamba variable `s` is the current element of the map.
The body of the lamba expresstion `"$acc\n$s"` is saying to take the `acc` variable, and append to it `\n' and `s`, then store that back into `acc` for the next iteration.

library/src/main/resources/SAMLSpecRefMessage.properties
```diff
@@ -160,6 +160,10 @@ SAMLCore_3_2_1_b=The identifier for the version of SAML defined in this specific
 
 SAMLCore_3_2_1_c=The time value is encoded in UTC, as described in Section 1.3.3.
 
+SAMLCore_3_2_1_d=If a SAML responder deems a request to be invalid according to SAML syntax or \
```
bakejeyner: ✏️ If we are passing in our own status code to test on, it may make sense to omit the last part of this reference:
`
If a SAML responder deems a request to be invalid according to SAML syntax or \
processing rules, then if it responds, it MUST return a SAML response message with a \
<StatusCode> element
`

test/common/src/main/kotlin/org/codice/compliance/utils/TestCommon.kt
```diff
@@ -38,6 +38,12 @@ class TestCommon {
                 "ItMustNotExceed80BytesInLength"
         const val MAX_RELAYSTATE_LEN = 80
 
+        const val REQUESTER = "urn:oasis:names:tc:SAML:2.0:status:Requester"
+        const val VERSION_MISMATCH = "urn:oasis:names:tc:SAML:2.0:status:VersionMismatch"
+        private const val SUCCESS = "urn:oasis:names:tc:SAML:2.0:status:Success"
```
coyotesqrl: ❓ Do you intend to use the `SUCCESS` and `RESPONDER` consts at a later date?

blen-desta: We will need `RESPONDER` for the SP tests and not sure right now about `SUCCESS`

PR #46, opened 2018-03-28
Removes the shouldBe checks on the status codes

bakejeyner:
Removes the shouldBe checks on the status codes and replaces the checks with our own in order to throw our own SAMLComplienceExceptions

### DIFF
--- a/test/common/src/main/kotlin/org/codice/compliance/verification/binding/BindingVerifier.kt
+++ b/test/common/src/main/kotlin/org/codice/compliance/verification/binding/BindingVerifier.kt
@@ -14,8 +14,8 @@
 package org.codice.compliance.verification.binding
 
 import com.google.api.client.http.HttpStatusCodes
+import org.codice.compliance.SAMLBindings_3_5_6_a
 import org.codice.compliance.SAMLComplianceException
-import org.codice.compliance.SAMLSpecRefMessage
 
 open class BindingVerifier {
     companion object {
@@ -28,7 +28,7 @@ open class BindingVerifier {
         fun verifyHttpStatusCode(code: Int) {
             if (code >= HTTP_ERROR_THRESHOLD) {
                 throw SAMLComplianceException.createWithPropertyMessage(
-                        SAMLSpecRefMessage.SAMLBindings_3_5_6_a,
+                        SAMLBindings_3_5_6_a,
                         property = "HTTP Status Code",
                         actual = code.toString(),
                         expected = "${HttpStatusCodes.STATUS_CODE_OK}"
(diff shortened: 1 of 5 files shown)


test/common/src/main/kotlin/org/codice/compliance/verification/binding/PostBindingVerifier.kt
```diff
@@ -41,13 +39,13 @@ import org.codice.compliance.utils.decorators.IdpPostResponseDecorator
 import org.codice.security.saml.SamlProtocol.Binding.HTTP_POST
 import org.codice.security.sign.Decoder
 
-class PostBindingVerifier(private val response: IdpPostResponseDecorator) {
+class PostBindingVerifier(private val response: IdpPostResponseDecorator) : BindingVerifier() {
```
coyotesqrl: ❓ What's the advantage to subclassing `BindingVerifier` here? We get no behaviors/polymorphism from it; we only use a static method from its companion.

test/common/src/main/kotlin/org/codice/compliance/verification/binding/BindingVerifier.kt
```diff
@@ -0,0 +1,39 @@
...
+         * according to the binding spec
+         * 3.4.6 & 3.5.6 Error Reporting
+         */
+        fun verifyHttpStatusCode(code: Int) {
+            if (code >= HTTP_ERROR_THRESHOLD) {
+                throw SAMLComplianceException.createWithPropertyMessage(
+                        SAMLSpecRefMessage.SAMLBindings_3_5_6_a,
+                        property = "HTTP Status Code",
+                        actual = code.toString(),
+                        expected = "${HttpStatusCodes.STATUS_CODE_OK}"
```
coyotesqrl: ❓ Other codes are acceptable but we're printing out an expected value of 200. Is that okay, or should be be more broad in our expected value message?

test/common/src/main/kotlin/org/codice/compliance/verification/binding/PostBindingVerifier.kt
brjeter: It was just so we could reuse the functionality. And we figure the reusable functionality that we need between POST and Redirect will grow.

test/common/src/main/kotlin/org/codice/compliance/verification/binding/BindingVerifier.kt
bakejeyner: Whoops. Forgot to change that when Jeter and I decided to check for >400. Fixing now.

PR #48, opened 2018-03-28
Printing the specific XML piece that's invalid when an exception occurs

idperez:


test/common/src/main/kotlin/org/codice/compliance/verification/binding/RedirectBindingVerifier.kt
```diff
@@ -177,7 +177,7 @@ class RedirectBindingVerifier(private val response: IdpRedirectResponseDecorator
      */
     private fun verifyNoXMLSig() {
         if (response.responseDom.children("Signature").isNotEmpty()) {
-            throw SAMLComplianceException.create(SAMLBindings_3_4_4_1, message = "Signature element found.")
+            throw SAMLComplianceException.create(SAMLBindings_3_4_4_1, message = "Signature element found.", node = response.responseDom)
```
bakejeyner: ❓ Is this line too long? Just making sure because it looks Strings are cut off early, like this one https://github.com/connexta/saml-conformance/pull/48/files#diff-9b82dc48c358096a39643f5081ef7f97R255

idperez: Build passes, `gradle spotlessApply` doesn't seem to mind it

idperez: https://github.com/connexta/saml-conformance/pull/49 this should resolve any line length issues

PR #51, opened 2018-03-29
Adds specific subclasses of the SAMLSpecRefMessage

coyotesqrl:
Includes doc names and URLs for the doc references.

Also fixes code trimming.


test/common/src/main/kotlin/org/codice/compliance/verification/core/CommonDataTypeVerifier.kt
```diff
@@ -153,7 +153,7 @@ private fun verifyYear(year: String, errorCode: SAMLSpecRefMessage?) {
     else arrayOf(errorCode)
 
     // check if year is an integer && https://www.w3.org/TR/xmlschema-2/#dateTime "a plus sign is not permitted"
-    if (!strippedYear.matches(Regex("\\d+"))) {
+    if (!strippedYear.matches("""\d+""".toRegex())) {
```
coyotesqrl: Small change just to be consistent in our use of regexes.

PR #54, opened 2018-03-29
Jenkinsfile typo

blen-desta:
😅


oconnormi:
What do they say about great minds?

PR #58, opened 2018-03-30
Added a depends on in docker copy task

blen-desta:
I noticed that when doing a parallel docker build, the Dockerfile fails to copy the distribution because the `copyDistribution` runs before it is built.


coyotesqrl:
Recommend closing this PR and letting @brjeter incorporate this change into his so neither of you needs to rebase after the fact.

blen-desta:
Sounds good. @brjeter ☝️

PR #63, opened 2018-04-02
Added two negative path tests from the profiles document

blen-desta:
Tests for 
```
Profiles 4.1.4.1:
If the identity provider cannot or will not satisfy the request, 
it MUST respond with a <Response> message containing an 
appropriate error status code or codes.
```
and
```
Profiles 4.1.4.1:
Note that the service provider MAY include a <Subject> element 
in the request that names the actual identity about which it wishes
to receive an assertion. This element MUST NOT contain any 
<SubjectConfirmation> elements. If the identity provider does
not recognize the principal as that identity, then it MUST respond 
with a <Response> message containing an error status and no assertions.
```


gradlew.bat
```diff
@@ -1,84 +1,84 @@
...
-rem Set variable GRADLE_EXIT_CONSOLE if you need the _script_ return code instead of
-rem the _cmd.exe /c_ return code!
-if  not "" == "%GRADLE_EXIT_CONSOLE%" exit 1
-exit /b 1
-
-:mainEnd
-if "%OS%"=="Windows_NT" endlocal
-
-:omega
+@if "%DEBUG%" == "" @echo off
```
blen-desta: The difference is in line endings

ctk/common/src/main/kotlin/org/codice/compliance/verification/profile/ProfilesVerifier.kt
```diff
@@ -31,11 +32,15 @@ class ProfilesVerifier(private val node: Node) {
      * Verify Error Response against the Profiles document.
      * This should be called explicitly if an error is expected.
      */
-    fun verifyErrorResponse() {
-        if (node.allChildren("Assertion").isNotEmpty())
-            throw SAMLComplianceException.create(SAMLProfiles_4_1_4_2_l,
+    fun verifyErrorResponseAssertion(samlErrorCode: SAMLSpecRefMessage? = null) {
+        if (node.allChildren("Assertion").isNotEmpty()) {
+            var exceptions: Array<SAMLSpecRefMessage> = arrayOf(SAMLProfiles_4_1_4_2_l)
```
coyotesqrl: ✏️ How about 
```kotlin
val exceptions: Array<SAMLSpecRefMessage> = 
  if (samlErrorCode == null) arrayOf(SAMLProfiles_4_1_4_2_l)
 else arrayOf(samlErrorCode, SAMLProfiles_4_1_4_2_l)
```
Could probably use a `var` and do a conditional append as well.

blen-desta: 👍 I like that

PR #71, opened 2018-04-03
Refactored SamlAssertionsVerifier

blen-desta:
`SamlAssertionsVerifier` is ridiculously large.

I split `SamlAssertionsVerifier` into:
- `AssertionsVerifier`
- `SubjectVerifier`
- `ConditionsVerifier`
- `StatementVerifier`

Note: Each of these is a section under `2 SAML Assertions`.

`SamlAssertionsVerifier` calls the `verify` method in each of the above classes so we don't have to call them individually. I left all these classes in the same file so the only benefit we get right now is clarity and not suppressing the static analysis finding from detekt. 

Should this be done or do you guys think we should leave it as is?

### DIFF
--- a/ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CommonDataTypeVerifier.kt
+++ b/ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CommonDataTypeVerifier.kt
@@ -35,11 +35,7 @@ const val FOUR_DIGIT_YEAR_LEN = 4
 
 var ids = mutableListOf<String>()
 
-/**
- * Verify common data types against the core specification
- *
- * 1.3 Common Data Types
- */
+/** 1.3 Common Data Types **/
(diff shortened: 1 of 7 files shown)


coyotesqrl:
@bakejeyner I'm not sure there's a lot to be gained by making them sealed. We're not defining an extended enum, which is the most common use case for sealed classes; we're just splitting work into meaningful chunks.

With a single file, we could make the classes private, though.

blen-desta:
- Created a package under `core` named `internal`
- Split and moved those classes under the `internal` package
- Made the public `verify` functions of each class `internal` and left the rest private since they are only called from the verify function in the particular class.

blen-desta:
One thing with making the classes internal:
```
internal declarations become public in Java. Members of internal classes go through 
name mangling, to make it harder to accidentally use them from Java and to allow 
overloading for members with the same signature that don't see each other according 
to Kotlin rules;
```

PR #74, opened 2018-04-05
Core 8.1.2 - Read/Write/Execute/Delete/Control with Negation

bakejeyner:
Adds a class `SamlIdenfiersVerifier` to lay the framework for all of section 8 testing.


ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlIdentifiersVerifier.kt
```diff
@@ -0,0 +1,56 @@
...
+        node.allChildren("Action").forEach({ action ->
+            action.nodeValue?.let {
+                checkActionNamespaceValue(it)
+            }
+        })
+    }
+
+    private fun checkActionNamespaceValue(nodeValue: String) {
+        actionNamespaces.forEach({
+            if (nodeValue.contains(it) && nodeValue.contains("~$it")) {
```
coyotesqrl: ❓ I'm confused. Are the values comma-delimited or something? Would you get "Read, ~Read, Write" for example?

bakejeyner: That, my friend, is a great question I too asked. The spec does not say how the values are delimited, nor how to include multiple values in general, just that you cannot have both the action and it's negated counterpart. So that must is the only thing I am testing.
`A SAML authority MUST NOT authorize both an action and its negated form`

SmithJosh: Looking at section 2.7.4, the <AuthzDecisionStatement> element can contain any number of <Action> elements. So perhaps there's no delimiters, and each is just a single action? Either way, it seems like we need to verify that there's no conflict between any <Action> element, not just inside a single element.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlIdentifiersVerifier.kt
```diff
@@ -0,0 +1,56 @@
...
+                throw SAMLComplianceException.create(
+                        codes = *arrayOf(SAMLCore_8_1_2),
```
blen-desta: ❓ I think you can just pass in `SAMLCore_8_1_2`.

blen-desta:
If it's for a request, isn't that an SP side test?

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlIdentifiersVerifier.kt
```diff
@@ -0,0 +1,77 @@
...
+        })
+
+        return actionList
+    }
+
+    private fun checkActionList(actionList: List<String>) {
+        RWEDC_LIST.forEach({
+            if (actionList.contains(it) && actionList.contains("~$it")) {
+                throw SAMLComplianceException.create(
+                        codes = SAMLCore_8_1_2,
```
coyotesqrl: ❗️ Listen to your IDE when it tells you something. 😉 
Remove `codes =` here so you do not rely on deprecated behavior.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlIdentifiersVerifier.kt
```diff
@@ -0,0 +1,77 @@
...
+                        message = """An "AuthzDecisionQuery" element contained an action and """ +
```
coyotesqrl: ✏️ There's no reason to use raw strings here.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlIdentifiersVerifier.kt
```diff
@@ -0,0 +1,77 @@
...
+            val actionList = createActionList(it)
+
+            if (actionList.isNotEmpty()) {
+                checkActionList(actionList)
+            }
+        })
+    }
+
+    private fun createActionList(query: Node): List<String> {
+        val actionList = mutableListOf<String>()
```
coyotesqrl: ✏️ Can simplify this function and remove the mutable list by
Firstly, getting rid of the two URI consts and replacing with
```kotlin
        private val RWEDC_URI_SET = setOf("urn:oasis:names:tc:SAML:1.0:action:rwedc-negation",
                "urn:oasis:names:tc:SAML:1.0:action:rwedc")
```
and then
```kotlin
        return query.children("Action")
                .filter { it.attributes.getNamedItem("Namespace").nodeValue in RWEDC_URI_SET }
                .map { it.nodeValue }
                .toList()
```

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlIdentifiersVerifier.kt
```diff
@@ -0,0 +1,77 @@
...
+            if (namespaceValue == RWEDC_URI || namespaceValue == RWEDC_NEGATION_URI) {
+                actionList.add(it.nodeValue)
+            }
+        })
+
+        return actionList
+    }
+
+    private fun checkActionList(actionList: List<String>) {
+        RWEDC_LIST.forEach({
```
coyotesqrl: ✏️ An alternate solution would remove the `RWEDC_LIST` and do the following:
```kotlin
        val (negated, notNegated) = actionList.partition { it.startsWith("~") }
        notNegated.forEach {
            if ("~$it" in negated) {
                throw SAMLComplianceException.create(
                        SAMLCore_8_1_2,
                        message = """An "AuthzDecisionQuery" element contained an action and """ +
                                """its negated form.""",
                        node = node
                )
            }
        }
```

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlIdentifiersVerifier.kt
```diff
@@ -0,0 +1,74 @@
...
+                .toList()
+    }
+
+    private fun checkActionList(actionList: List<String>) {
+        val (negated, notNegated) = actionList.partition { it.startsWith("~") }
+        notNegated.forEach {
+            if ("~$it" in negated) {
+                throw SAMLComplianceException.create(
+                        SAMLCore_8_1_2,
+                        message = """An "AuthzDecisionQuery" element contained an action and """ +
```
coyotesqrl: ✏️ Still would remove the raw strings here as they add no value.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlIdentifiersVerifier.kt
```diff
@@ -0,0 +1,74 @@
...
+import org.w3c.dom.Node
+
+internal class SamlIdentifiersVerifier(val node: Node) {
+    companion object {
+        private val RWEDC_URI_SET = setOf(
+                "urn:oasis:names:tc:SAML:1.0:action:rwedc-negation",
+                "urn:oasis:names:tc:SAML:1.0:action:rwedc"
+        )
+
+        private val RWEDC_LIST = listOf(
```
coyotesqrl: ❗️ Unused.

PR #76, opened 2018-04-05
MAINT - Stops detekt from excluding files with names ending in `Test` or `Spec`

coyotesqrl:


build.gradle
```diff
@@ -92,7 +92,6 @@ build {
 
 detekt {
     version = "$detekt_version"
-    profile = 'main'
```
brjeter: Obscure 😆

detekt.yml
```diff
@@ -34,6 +34,9 @@ output-reports:
   #  - 'PlainOutputReport'
   #  - 'XmlOutputReport'
 
+test-pattern:
+  active: false
```
brjeter: I thought there would some sort of property we could set for this 👍

PR #82, opened 2018-04-09
KEYCLOAK - Fixed tests to pass against Keycloak

brjeter:
- Debug whole HTTP body on POST response
 - Only check relay state if it's given
 - Fix bug in pretty printing with ampersands

### DIFF
--- a/ctk/common/src/main/kotlin/org/codice/compliance/utils/decorators/IdpPostResponseDecorator.kt
+++ b/ctk/common/src/main/kotlin/org/codice/compliance/utils/decorators/IdpPostResponseDecorator.kt
@@ -15,6 +15,7 @@ package org.codice.compliance.utils.decorators
 
 import org.apache.cxf.rs.security.saml.sso.SSOConstants.RELAY_STATE
 import org.apache.cxf.rs.security.saml.sso.SSOConstants.SAML_RESPONSE
+import org.codice.compliance.debugPrettyPrintXml
 import org.codice.compliance.saml.plugin.IdpPostResponse
 import org.codice.compliance.utils.TestCommon.Companion.acsUrl
 import org.codice.compliance.verification.binding.BindingVerifier
@@ -37,6 +38,10 @@ internal constructor(response: IdpPostResponse) : IdpPostResponse(response), Idp
         private const val POST = "POST"
     }
 
+    init {
+        responseBodyString?.debugPrettyPrintXml("HTTP Response Body")
+    }
+
     override var isRelayStateGiven: Boolean = false
     override lateinit var decodedSamlResponse: String
 
@@ -49,14 +54,14 @@ internal constructor(response: IdpPostResponse) : IdpPostResponse(response), Idp
         responseForm == null
     }
     val isSamlResponseFormNull: Boolean by lazy {
-        samlResponseForm == null
+        samlResponseFormControl == null
     }
     val isRelayStateFormNull: Boolean by lazy {
-        relayStateForm == null
+        relayStateFormControl == null
     }
 
     val isSamlResponseNameCorrect: Boolean by lazy {
-        checkNodeAttribute(samlResponseForm, NAME, SAML_RESPONSE)
+        checkNodeAttribute(samlResponseFormControl, NAME, SAML_RESPONSE)
     }
 
     /*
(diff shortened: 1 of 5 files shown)


ctk/common/src/main/kotlin/org/codice/compliance/utils/decorators/IdpPostResponseDecorator.kt
```diff
@@ -49,14 +54,14 @@ internal constructor(response: IdpPostResponse) : IdpPostResponse(response), Idp
         responseForm == null
     }
     val isSamlResponseFormNull: Boolean by lazy {
-        samlResponseForm == null
+        samlResponseFormControl == null
     }
     val isRelayStateFormNull: Boolean by lazy {
-        relayStateForm == null
+        relayStateFormControl == null
```
brjeter: Renamed the `relayState` and `samlResponse` fields for clarity.

ctk/common/src/main/kotlin/org/codice/compliance/verification/profile/SingleSignOnProfileVerifier.kt
```diff
@@ -100,7 +102,9 @@ class SingleSignOnProfileVerifier(private val response: Node, private val acsUrl
      */
     private fun verifySsoAssertions() {
         val assertions = response.children("Assertion")
-        if (assertions.isEmpty()) {
+        val encryptedAssertions = response.children("EncryptedAssertion")
+
+        if (assertions.isEmpty() && encryptedAssertions.isEmpty()) {
```
brjeter: @bakejeyner will fix this to actually decrypt the encrypted assertion and verify them like the other assertion elements.

library/src/main/kotlin/org/codice/compliance/Common.kt
```diff
@@ -162,19 +164,30 @@ fun Node.prettyPrintXml(): String {
...
-    }
+    // Escape all ampersands because Keycloak does not properly escape it in POST responses
+    // which causes the transform to fail.
+    val escapedString = this.replace(Regex("&([^;]+(?!(?:\\\\w|;)))"),
+            { match -> "&amp;${match.value.removePrefix("&")}" })
+    val input = StreamSource(StringReader(escapedString))
+    val output = StreamResult(StringWriter())
+    val transformer = createTransformer()
+    transformer.transform(input, output)
+    return output.writer.toString()
```
brjeter: @coyotesqrl Your last commit made it so HTTP response bodies were printed all on one line for some reason. If you know why off the top of your head I can put the fix in this PR. Otherwise, I'm just reverting for String for now so that we can merge this PR.

library/src/main/kotlin/org/codice/compliance/Common.kt
```diff
@@ -162,19 +164,30 @@ fun Node.prettyPrintXml(): String {
...
 }
 
 fun String.debugPrettyPrintXml(header: String?) {
     Log.debugWithSupplier {
-        val headerVal = if (header == null) "$header:\n\n" else ""
+        val headerVal = if (header != null) "$header:\n\n" else ""
```
brjeter: Oops.

external/api/src/main/java/org/codice/compliance/saml/plugin/IdpPostResponse.java
```diff
@@ -51,21 +52,25 @@ public IdpPostResponse(Response response) {
 
     if (responseForm != null) {
       parseAndSetFormValues();
+    } else {
+      // Purely for debugging purposes
+      responseBodyString = response.then().extract().body().asString();
```
coyotesqrl: ✏️ While this isn't a super-expensive operation, it might be worth doing lazy initialization of it (harder in Java, I know) if it's ever invoked. Maybe an AtomicReference with a method to fill it? Not sure.

library/src/main/kotlin/org/codice/compliance/Common.kt
coyotesqrl: Don't know why it would do that. It definitely should not have. I'll look after this is merged.

external/api/src/main/java/org/codice/compliance/saml/plugin/IdpPostResponse.java
brjeter: Well I'd have to save the RestAssured Response object (named `response` here) to a field. But that would work if I just save it to a field and then extract the body the one time I need to debug it. I'll give that a try.

PR #84, opened 2018-04-09
MAINT - Removed all the StringLiteralDuplication suppressions and addressed issues

blen-desta:
- Changed CommonDataTypeVerifier to a class with static methods
- Removed all the StringLiteralDuplication suppressions and addressed issues

### DIFF
--- a/ctk/common/src/main/kotlin/org/codice/compliance/utils/decorators/IdpRedirectResponseDecorator.kt
+++ b/ctk/common/src/main/kotlin/org/codice/compliance/utils/decorators/IdpRedirectResponseDecorator.kt
@@ -23,7 +23,7 @@ import org.codice.compliance.verification.binding.RedirectBindingVerifier
 import org.w3c.dom.Node
 
 /**
- * This class  can only be instantiated by using extension methods in IdpResponseDecorator.kt
+ * This class can only be instantiated by using extension methods in IdpResponseDecorator.kt
  */
 class IdpRedirectResponseDecorator
 internal constructor(response: IdpRedirectResponse) : IdpRedirectResponse(response),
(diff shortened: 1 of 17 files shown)


ctk/common/src/main/kotlin/org/codice/compliance/utils/decorators/IdpRedirectResponseDecorator.kt
```diff
@@ -50,17 +50,6 @@ internal constructor(response: IdpRedirectResponse) : IdpRedirectResponse(respon
         paramMap[SIGNATURE]
     }
 
-    val extraUrlParameters: Boolean by lazy {
```
blen-desta: Removed this unused method.

coyotesqrl: I'd say even if we keep it, there are better ways of determining its value.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CommonDataTypeVerifier.kt
```diff
@@ -31,156 +30,170 @@ import java.net.URI
...
+        private const val NO_HYPHEN_MESSAGE = "No hyphen was found in the date value of"
+
+        var ids = mutableListOf<String>()
+
+        /** 1.3 Common Data Types **/
+        fun verifyCommonDataType(samlDom: Node) {
+            ids = mutableListOf()
+            var i = samlDom.childNodes.length - 1
+
+            while (i >= 0) {
```
coyotesqrl: ✏️ A ranged loop would be cleaner:
`for (i in samlDom.childNodes.length - 1 downTo 0) {`
you can remove the var and get rid of the decrement on line 59 with that.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CommonDataTypeVerifier.kt
```diff
@@ -31,156 +30,170 @@ import java.net.URI
...
+                    verifyIdValues(child, null)
+
+                if (child.hasChildNodes())
+                    verifyCommonDataType(child)
+                i -= 1
+            }
+        }
+
+        /** 1.3.1 String Values **/
+        @SuppressWarnings("StringLiteralDuplication")
```
coyotesqrl: ❗️ Should be kotlin's `@Suppress` annotation.

PR #85, opened 2018-04-09
MAINT - Removed TODOs

blen-desta:
- Removed TODOs that are not relevant
- Removed TODOs in exception codes (since no one is looking at the existing one and we have tickets tracking them)
- Created cards for remaining TODOs

### DIFF
--- a/ctk/common/src/main/kotlin/org/codice/compliance/utils/TestCommon.kt
+++ b/ctk/common/src/main/kotlin/org/codice/compliance/utils/TestCommon.kt
@@ -122,15 +122,9 @@ class TestCommon {
          * test.
          * @return An {@code IdpResponse} object created from the error response.
          */
+        // TODO Change HTTP status code to expect 302/303
         fun parseErrorResponse(response: Response): IdpResponseDecorator {
             return if (response.header("Location") != null) {
-                /*
-                 * TODO "Manually change DDF IdP to respond with 302/303 status code for Redirect"
-                 * Change this line to:
-                 IdpRedirectResponse(response)
-                 *
-                 * And delete parseRedirectErrorResponse method when finished with ticket
-                 */
                 parseRedirectErrorResponse(response).decorate()
             } else {
                 IdpPostResponse(response).decorate()
--- a/ctk/common/src/main/kotlin/org/codice/compliance/verification/binding/RedirectBindingVerifier.kt
+++ b/ctk/common/src/main/kotlin/org/codice/compliance/verification/binding/RedirectBindingVerifier.kt
@@ -93,8 +93,7 @@ class RedirectBindingVerifier(private val response: IdpRedirectResponseDecorator
      * 3.4.6 Error Reporting
      */
     private fun verifyHttpRedirectStatusCode() {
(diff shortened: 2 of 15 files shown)


ctk/common/src/main/kotlin/org/codice/compliance/utils/TestCommon.kt
```diff
@@ -120,13 +120,6 @@ class TestCommon {
          */
         fun parseErrorResponse(response: Response): IdpResponseDecorator {
             return if (response.header("Location") != null) {
-                /*
-                 * TODO "Manually change DDF IdP to respond with 302/303 status code for Redirect"
```
bakejeyner: ❓ This TODO was to remind us that when DDF is fixed to return the correct HTTP status code. What was the reasoning behind taking it out?

ctk/common/src/main/kotlin/org/codice/compliance/verification/binding/RedirectBindingVerifier.kt
```diff
@@ -93,8 +93,6 @@ class RedirectBindingVerifier(private val response: IdpRedirectResponseDecorator
      * 3.4.6 Error Reporting
      */
     private fun verifyHttpRedirectStatusCode() {
-        // TODO remove the 200 check when "Manually change DDF IdP to respond with 302/303 status
```
bakejeyner: ❓ See above question.

ctk/common/src/main/kotlin/org/codice/compliance/verification/binding/RedirectBindingVerifier.kt
```diff
@@ -114,8 +112,6 @@ class RedirectBindingVerifier(private val response: IdpRedirectResponseDecorator
      * 3.4.6 Error Reporting
      */
     private fun verifyHttpRedirectStatusCodeErrorResponse() {
-        // TODO remove the 200 check when "Manually change DDF IdP to respond with 302/303 status
```
bakejeyner: ❓ See above question.

ctk/common/src/main/kotlin/org/codice/compliance/utils/TestCommon.kt
blen-desta: I removed it cause we aren't doing that anymore. We will be responding exclusively with POST. We still might want to do it for log out so I will document that in the ticket.

ctk/common/src/main/kotlin/org/codice/compliance/verification/binding/RedirectBindingVerifier.kt
bakejeyner: ❗️ Either leave a comment about allowing the 200 response code, or take out the extra check (`response.httpStatusCode != HttpStatusCodes.STATUS_CODE_OK`)

ctk/common/src/main/kotlin/org/codice/compliance/verification/binding/RedirectBindingVerifier.kt
bakejeyner: ❗️ See above exclamation.

PR #87, opened 2018-04-10
MAINT - Changes prettyPrintXml functionality to use the HTML output method

coyotesqrl:
This should resolve formatting/indenting for both xml and html input.


coyotesqrl:
build now

PR #90, opened 2018-04-11
Add SAML schema validation for IdP responses

SmithJosh:
Added a schema validator which should be able to be used for any SAML messages. Right now it's just checking the IdP responses after authenticating.

The schema validation check adds half a second or so to each test.


ctk/common/src/main/kotlin/org/codice/compliance/utils/schema/SchemaValidator.kt
```diff
@@ -0,0 +1,111 @@
...
+
+    private class SAMLErrorHandler : ErrorHandler {
+
+        private fun getParseExceptionInfo(spe: SAXParseException): String {
+            var systemId: String? = spe.systemId
+            if (systemId == null) {
+                systemId = "null"
+            }
+
+            return "URI=" + systemId + " Line=" + spe.lineNumber + ": " + spe.message
```
brjeter: ✏️ In Kotlin you can use [String Templates](https://kotlinlang.org/docs/reference/basic-types.html#string-templates) instead of concatenating Strings like this.
✏️ Also you can use the [Elvis Operator](https://kotlinlang.org/docs/reference/null-safety.html#elvis-operator)  to evaluate what `systemId` should be in the null case.

With both of the above, this method could be reduced to one line:
`return "URI=${spe.systemId ?: "null"} Line=${spe.lineNumber}: ${spe.message}"`

ctk/common/src/main/kotlin/org/codice/compliance/utils/schema/SchemaValidator.kt
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        @Throws(SAXException::class)
+        override fun warning(spe: SAXParseException) {
+            Log.debugWithSupplier { "Warning: " + getParseExceptionInfo(spe) }
```
brjeter: Do you know what an example of a warning would be in this case? We want to be pretty strict about them following the schema.

SmithJosh:
Updated the validator based on our discussion to provide more helpful output

ctk/idp/src/main/resources/saml.xml
```diff
@@ -0,0 +1,118 @@
+<saml2p:Response xmlns:saml2p="urn:oasis:names:tc:SAML:2.0:protocol" xmlns:xsd="http://www.w3.org/2001/XMLSchema" Destination="https://samlhost:8993/services/saml/sso/post" ID="_fc259ded-3d6e-402a-a74a-365c23a1782d" InResponseTo="a1chfeh0234hbifc1jjd3cb40ji0d49" IssueInstant="2018-04-10T17:03:50.642Z" Version="2.0">
```
brjeter: ❓ Did you mean to add this file? I'm assuming you used it for testing, but we probably don't need it in the project.

SmithJosh: Nope, it was just for testing. I'll remove it

SmithJosh:
build now

blen-desta:
☝️That won't work until https://github.com/connexta/saml-conformance/pull/93 gets merge. Wait for one more approval

SmithJosh:
build now

blen-desta:
build now

PR #92, opened 2018-04-11
Core - Node decryption

bakejeyner:
Adds a node decryption class

### DIFF
--- a/ctk/common/src/main/kotlin/org/codice/compliance/utils/XMLDecryptor.kt
+++ b/ctk/common/src/main/kotlin/org/codice/compliance/utils/XMLDecryptor.kt
@@ -17,7 +17,7 @@ import org.apache.wss4j.common.crypto.JasyptPasswordEncryptor
 import org.apache.wss4j.common.crypto.Merlin
 import org.apache.xml.security.encryption.XMLCipher
 import org.bouncycastle.jce.provider.BouncyCastleProvider
-import org.codice.compliance.allChildren
+import org.codice.compliance.recursiveChildren
 import org.codice.compliance.children
 import org.codice.compliance.utils.TestCommon.Companion.KEYSTORE_PASSWORD
 import org.codice.compliance.utils.TestCommon.Companion.PRIVATE_KEY_ALIAS
@@ -88,7 +88,7 @@ class XMLDecryptor {
             val ownerDocument = node.ownerDocument
             val encData = encElement.children("EncryptedData").first() as Element
 
-            val encKeyElements = encElement.allChildren("EncryptedKey")
+            val encKeyElements = encElement.recursiveChildren("EncryptedKey")
                     .map { it as Element }
             if (encKeyElements.isEmpty()) messageBuilder.append("No Keys could be found.\n")
 
--- a/ctk/common/src/main/kotlin/org/codice/compliance/utils/XMLDecryptor.kt
+++ b/ctk/common/src/main/kotlin/org/codice/compliance/utils/XMLDecryptor.kt
@@ -1,0 +1,163 @@
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.compliance.utils
+
+import org.apache.wss4j.common.crypto.JasyptPasswordEncryptor
+import org.apache.wss4j.common.crypto.Merlin
+import org.apache.xml.security.encryption.XMLCipher
+import org.bouncycastle.jce.provider.BouncyCastleProvider
+import org.codice.compliance.allChildren
+import org.codice.compliance.children
+import org.w3c.dom.Document
+import org.w3c.dom.Element
+import org.w3c.dom.Node
+import java.io.File
+import java.security.Key
+import java.security.Security
+import java.util.Properties
+
+class XMLDecryptor {
+    companion object {
+        // One time initialization
+        init {
+            Security.addProvider(BouncyCastleProvider())
+            org.apache.xml.security.Init.init()
+        }
+
+        private const val ENCRYPTION_FILE_NAME = "encryption.properties"
+        private const val BC_PROVIDER = "BC"
+
+        private val serverPrivateKey by lazy {
+
+            val encryptionFile = File(ClassLoader.getSystemResource(ENCRYPTION_FILE_NAME).toURI())
+            checkNotNull(encryptionFile)
+
+            val encryptionProperties = encryptionFile.inputStream().use {
+                Properties().apply {
+                    load(it)
+                }
+            }
+
+            val keystorePassword = encryptionProperties
+                    .getProperty("org.apache.ws.security.crypto.merlin.keystore.password")
+            val keystorePrivateKeyPassword = encryptionProperties
+                    .getProperty("org.apache.ws.security.crypto.merlin.keystore.private.password")
+            val keystorePrivateKeyAlias = encryptionProperties
+                    .getProperty("org.apache.ws.security.crypto.merlin.keystore.alias")
+
+            // Using Apache Merlin to read the keystore
+            val merlin = Merlin(encryptionProperties,
+                    ClassLoader.getSystemClassLoader(),
+                    JasyptPasswordEncryptor(keystorePassword))
+
+            merlin.getPrivateKey(keystorePrivateKeyAlias, keystorePrivateKeyPassword)
+        }
+
+        /**
+         * Takes in an EncryptedID, EncryptedAssertion, or EncryptedAttribute element and returns
+         * the decrypted node.
+         *
+         * Only supports up to one level encryption using the EncryptedKey element. In other words,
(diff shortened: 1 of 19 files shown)


ctk/common/src/main/kotlin/org/codice/compliance/utils/XMLDecryptor.kt
```diff
@@ -0,0 +1,163 @@
...
+         * the data must be directly encrypted by the CTK's public key, or, if another key is used
+         * to encrypt the data, that key must be encrypted using the CTK's public key and put into
+         * the EncryptedData element.
+         *
+         * @param node An encrypted {@code Node} to be decrypted
+         * @return Node The decrypted {@code Node}
+         */
+        @Suppress("TooGenericExceptionCaught"
+                /* Decryption error handling is independent of Exception type */)
+        fun decryptAndReplaceNode(node: Node): Node {
```
brjeter: ✏️ I'd recommend either making this function `internal` so that only files in `ctk/common` can access it OR having a require statement requiring that `node.localName` is equal to one of the 3 possible encrypted elements. Maybe do both.

ctk/common/src/main/kotlin/org/codice/compliance/utils/XMLDecryptor.kt
```diff
@@ -0,0 +1,163 @@
...
+                // Keep a reference Node, either a sibling (if it has one) or parent, of `node` in
+                // order to return the decrypted version of `node`
+                return if (encData.previousSibling != null) {
+                    val referenceNode = encData.previousSibling
+                    encryptionCipher.doFinal(ownerDocument, encData)
+                    replaceNode(referenceNode.nextSibling, node)
+                } else {
+                    val referenceNode = encData.parentNode
+                    encryptionCipher.doFinal(ownerDocument, encData)
+                    replaceNode(referenceNode.firstChild, node)
```
brjeter: ✏️ I still don't like that you're doing this weird replace based on either the sibling or the parent. I liked the way Keycloak was doing it with using the returned document from `encryptionCipher.doFinal()` and replacing from the root node.

ctk/common/src/main/kotlin/org/codice/compliance/utils/XMLDecryptor.kt
```diff
@@ -0,0 +1,163 @@
...
+
+            encKeyElements.forEach {
+                val encryptedKey = cipher.loadEncryptedKey(ownerDocument, it)
+
+                try {
+                    return keyCipher.decryptKey(
+                            encryptedKey,
+                            encryptedKey.encryptionMethod.algorithm)
+                } catch (e: Exception) {
+                    // Continue the loop. Keep trying. Don't give up.
```
brjeter: "Don't give up" by Eagle-eye Cherry
https://youtu.be/ziDG8bBqp38

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/EncryptionVerifier.kt
```diff
@@ -0,0 +1,71 @@
...
+     * <li>Decrypting the elements</li>
+     * <li>Replacing the encrypted elements with their unencrypted values</li>
+     * </ol>
+     *
+     * @param response The response node to verify and decrypt
+     */
+    fun verifyAndDecrypt(response: Node) {
+        val encryptedNodeList = response.allChildren("EncryptedAssertion")
+                .union(response.allChildren("EncryptedAttribute"))
+                .union(response.allChildren("EncryptedId"))
```
brjeter: ❗️ the "d" in "EncryptedId" should be capitalized.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/EncryptionVerifier.kt
```diff
@@ -0,0 +1,71 @@
...
+
+        encryptedNodeList.forEach({
+            verifyEncryptedElement(it)
```
brjeter: Could you make everything inside this for loop another public function that you just call? I'm finding that I need to do these operations on an individual element for the NameIDPolicy stuff I'm working on.

ctk/common/src/main/kotlin/org/codice/compliance/utils/XMLDecryptor.kt
```diff
@@ -0,0 +1,163 @@
...
+            checkNotNull(encryptionFile)
+
+            val encryptionProperties = encryptionFile.inputStream().use {
+                Properties().apply {
+                    load(it)
+                }
+            }
+
+            val keystorePassword = encryptionProperties
+                    .getProperty("org.apache.ws.security.crypto.merlin.keystore.password")
```
blen-desta: ✏️ You could change `org.apache.ws.security.crypto.merlin.keystore.password` etc. into constants. I think they are used in library

ctk/common/src/main/kotlin/org/codice/compliance/utils/XMLDecryptor.kt
bakejeyner: 👍  I have an idea that may make it look cleaner. If that ends up being a bust I'll look back at Keycloak's solution.

ctk/common/src/main/kotlin/org/codice/compliance/utils/XMLDecryptor.kt
bakejeyner: The javadoc mentions the allowable encrypted elements.

ctk/common/src/main/kotlin/org/codice/compliance/utils/XMLDecryptor.kt
bakejeyner: My idea didn't plan out. And as for Keycloak, they are making certain assumptions that we can't. They assume that the first child of the root element is the encrypted element. I'm not sure if this is because they put the encrypted element into a new document before sending it to the XML processing function, or if they just make bad assumptions, but either way we can't assume that for our purposes. Any other ideas are welcome though!

Keycloack reference: https://github.com/keycloak/keycloak/blob/17c13043d0118ecb866c294e32c30a7bcb8ddb13/saml-core/src/main/java/org/keycloak/saml/processing/core/util/XMLEncryptionUtil.java#L378

brjeter: Yeah you're right. They create a brand new document and pass that into the function... I don't have any better ideas.

bakejeyner:
build now

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/EncryptionVerifier.kt
```diff
@@ -66,7 +68,12 @@ class EncryptionVerifier {
                         .attributes.getNamedItem("Type")
                         .textContent != TestCommon.ELEMENT)
             throw SAMLComplianceException.createWithPropertyMessage(
-                    SAMLCore_2_7_3_2_a,
+                    when (encryptedElement.localName) {
+                        "EncryptedID" -> SAMLCore_2_2_4_a
+                        "EncryptedAssertion" -> SAMLCore_2_3_4_a
+                        "EncryptedAttribute" -> SAMLCore_2_7_3_2_a
+                        else -> throw UnknownError("Unknown ${encryptedElement.localName} type.")
+                    },
```
brjeter: 💯

bakejeyner:
build now

ctk/common/src/main/kotlin/org/codice/compliance/utils/XMLDecryptor.kt
```diff
@@ -0,0 +1,160 @@
...
+                    ?: serverPrivateKey
+            if (encryptionKey == serverPrivateKey)
+                messageBuilder.append("No Keys could be decrypted.\n")
+
+            val encryptionCipher = XMLCipher.getProviderInstance(BC_PROVIDER).apply {
+                init(XMLCipher.DECRYPT_MODE, encryptionKey)
+            }
+
+            try {
+                // Keep a reference Node, either a sibling (if it has one) or parent, of `node` in
```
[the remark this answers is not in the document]
bakejeyner: The problem is that the parent node may have multiple children, and the encrypted value may not be the first child. We would need to find and keep track of the child in that case.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/EncryptionVerifier.kt
```diff
@@ -0,0 +1,82 @@
...
+     * <li>Locating all of the encrypted elements within the given response</li>
+     * <li>Running verifications on the elements</li>
+     * <li>Decrypting the elements</li>
+     * <li>Replacing the encrypted elements with their unencrypted values</li>
+     * </ol>
+     *
+     * @param response The response node to verify and decrypt
+     */
+    fun verifyAndDecryptResponse(response: Node) {
+        val encryptedNodeList = response.allChildren("EncryptedAssertion")
```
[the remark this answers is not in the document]
bakejeyner: Just to make sure that we have no repeated elements (which would cause a false exception to be thrown), I changed
```
response.allChildren("EncryptedID")).forEach {
```
to
```
response.allChildren("EncryptedID")).toSet().forEach {
```

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/EncryptionVerifier.kt
```diff
@@ -0,0 +1,82 @@
...
+     * <li>Decrypting the elements</li>
+     * <li>Replacing the encrypted elements with their unencrypted values</li>
+     * </ol>
+     *
+     * @param response The response node to verify and decrypt
+     */
+    fun verifyAndDecryptResponse(response: Node) {
+        sequenceOf(response.allChildren("EncryptedAssertion"),
+                response.allChildren("EncryptedAttribute"),
+                response.allChildren("EncryptedID")).toSet().forEach {
```
coyotesqrl: This isn't doing what you think it's doing.

bakejeyner:
build now

bakejeyner:
build now

PR #95, opened 2018-04-12
Core - Added verification for SAML extensions

brjeter:
Verify for SAML extensions for both `<Attribute>` and `<SubjectConfirmationData>` elements. Here's the spec snippet which is identical in both:
```
SAML extensions MUST NOT add local (non-namespace-qualified) XML attributes or XML 
attributes qualified by a SAML-defined namespace to the SubjectConfirmationDataType 
complex type or a derivation of it; such attributes are reserved for future maintenance 
and enhancement of SAML itself.
```
#### Checklist:
☝️ Had that updated *before* I got the PR up BTW! 😄

### DIFF
--- a/ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
+++ b/ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
@@ -21,17 +21,51 @@ import org.codice.compliance.SAMLCore_5_4_2_a
 import org.codice.compliance.SAMLCore_5_4_2_b
 import org.codice.compliance.SAMLCore_5_4_2_b1
 import org.codice.compliance.SAMLCore_6_1_b
+import org.codice.compliance.SAMLCore_SamlExtensions
 import org.codice.compliance.SAMLSpecRefMessage
 import org.codice.compliance.allChildren
 import org.codice.compliance.children
+import org.codice.compliance.utils.TestCommon
 import org.codice.compliance.utils.TestCommon.Companion.ELEMENT
 import org.codice.compliance.utils.TestCommon.Companion.REQUESTER
 import org.codice.compliance.verification.core.CommonDataTypeVerifier.Companion.verifyCommonDataType
+import org.w3c.dom.Attr
+import org.w3c.dom.Element
 import org.w3c.dom.Node
 
 class CoreVerifier(val node: Node) {
     companion object {
         private const val ENCRYPTED_DATA = "EncryptedData"
+
+        /**
+         * Verify SAML extension attributes or elements against the Core Spec document
+         *
+         * 2.4.1.2 Element <SubjectConfirmationData>
+         * 2.7.3.1 Element <Attribute>
+         * 3.2.1 Complex Type RequestAbstractType
+         * 3.2.2 Complex Type StatusResponseType
+         */
+        internal fun verifySamlExtensions(nodes: List<Node>,
+                                          expectedSamlNames: List<String>) {
+            nodes.forEach {
+                if (isNullNamespace(it) || (it.namespaceURI == TestCommon.SAML_NAMESPACE
+                                && !expectedSamlNames.contains(it.localName))) {
+                    throw SAMLComplianceException.create(SAMLCore_SamlExtensions,
+                            message = "An invalid SAML extension was found.",
+                            node = it)
+                }
+            }
+        }
+
+        private fun isNullNamespace(node: Node): Boolean {
+            return with (node) {
+                when (this) {
+                    is Attr -> namespaceURI == null && ownerElement.namespaceURI == null
+                    is Element -> namespaceURI == null
+                    else -> throw UnknownError("Unknown Node type found")
+                }
+            }
+        }
     }
 
     /**
--- a/ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
+++ b/ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
@@ -48,7 +48,7 @@ class CoreVerifier(val node: Node) {
         internal fun verifySamlExtensions(nodes: List<Node>,
                                           expectedSamlNames: List<String>) {
             nodes.forEach {
-                if (isNullNamespace(it) || (it.namespaceURI == TestCommon.SAML_NAMESPACE
+                if (isNullNamespace(it) || (isSamlNamespace(it)
                                 && !expectedSamlNames.contains(it.localName))) {
                     throw SAMLComplianceException.create(SAMLCore_SamlExtensions,
                             message = "An invalid SAML extension was found.",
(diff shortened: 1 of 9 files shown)


brjeter:
I decided to just duplicate the logic once instead of making a common function since it only has and will only ever have two usages.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/internal/StatementVerifier.kt
```diff
@@ -84,8 +86,10 @@ internal class StatementVerifier(val node: Node) {
 
             if (encryptedData
```
bakejeyner: ❗️ Since I am removing this check in my PR it may be better to either take this out completely or leave it alone.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/internal/StatementVerifier.kt
```diff
@@ -113,9 +118,30 @@ internal class StatementVerifier(val node: Node) {
                     || (friendlyNameAttr != null && friendlyNameAttr.textContent == null)) {
                 verifyAttributeValue(it)
             }
+
+            for (i in it.attributes.length - 1 downTo 0) {
```
bakejeyner: ✏️ Sort-of related to the above comment: In an upcoming PR I'm adding an `attributeList()` extension function to the `Node` class in `Common.kt`.

```
fun Node.attributeList(): List<Node> {
    val attributesList = mutableListOf<Node>()
    this.attributes.let {
        for (i in it.length downTo 0) {
            attributesList.add(it.item(i))
        }
    }
    return attributesList
}
```
If you want you can add that now and stream the attributeList instead.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/internal/SubjectVerifier.kt
```diff
@@ -71,6 +72,28 @@ internal class SubjectVerifier(val node: Node) {
                 throw SAMLComplianceException.create(SAMLCore_2_4_1_3, XMLSignature_4_5,
                         message = "Multiple Keys found within the KeyInfo element.",
                         node = node)
+
+            for (i in it.attributes.length - 1 downTo 0) {
```
bakejeyner: ✏️ See above comment.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/internal/StatementVerifier.kt
brjeter: I'm technically not changing any functionality here. I guess the formatter must've kicked in. I could take this out.

library/src/main/kotlin/org/codice/compliance/Common.kt
```diff
@@ -118,11 +118,11 @@ fun Log.debugWithSupplier(message: () -> String) {
  * @param name - Name of Assertions.children
  * @return list of Assertions.children matching the name provided
  */
-fun Node.children(name: String): List<Node> {
+fun Node.children(name: String = "default"): List<Node> {
     val childNodes = mutableListOf<Node>()
     for (i in (this.childNodes.length - 1) downTo 0) {
         this.childNodes.item(i).apply {
-            if (localName == name) childNodes.add(this)
+            if (localName == name || name == "default") childNodes.add(this)
```
brjeter: Open to ideas on how to make this cleaner. I want to provide the option of not passing in the `name` parameter and just returning all children.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
```diff
@@ -32,6 +34,24 @@ import org.w3c.dom.Node
...
+         */
+        internal fun verifySamlExtensions(nodes: List<Node>) {
+            nodes.forEach {
+                if (it.namespaceURI == null || it.namespaceURI == TestCommon.SAML_NAMESPACE) {
+                    throw SAMLComplianceException.create(SAMLCore_SamlExtensions,
+                            message = "An invalid SAML extension was found.",
+                            node = it)
+                }
+            }
+        }
```
brjeter: Open to feedback on a better place to put this. Wasn't sure where I should do that.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
```diff
@@ -32,6 +34,24 @@ import org.w3c.dom.Node
...
+
+        /**
+         * Verify SAML extension attributes or elements against the Core Spec document
+         *
+         * 2.4.1.2 Element <SubjectConfirmationData>
+         * 2.7.3.1 Element <Attribute>
+         * 3.2.1 Complex Type RequestAbstractType
+         * 3.2.2 Complex Type StatusResponseType
+         */
+        internal fun verifySamlExtensions(nodes: List<Node>) {
```
brjeter: Open to feedback on a better place to put this. Wasn't sure where I should do that.

brjeter:
build now

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
```diff
@@ -21,17 +21,64 @@ import org.codice.compliance.SAMLCore_5_4_2_a
...
+         * Verify SAML extension attributes or elements against the Core Spec document
+         *
+         * 2.4.1.2 Element <SubjectConfirmationData>
+         * 2.7.3.1 Element <Attribute>
+         * 3.2.1 Complex Type RequestAbstractType
+         * 3.2.2 Complex Type StatusResponseType
+         */
+        internal fun verifySamlExtensions(nodes: List<Node>,
+                                          expectedSamlNames: List<String>) {
+            nodes.forEach {
```
bakejeyner: ✏️ Streaming alternative (I personally prefer the foreach):
```
if (nodes.stream().anyMatch {
            isNullNamespace(it) ||
                    (isSamlNamespace(it) && !expectedSamlNames.contains(it.localName))
        })
    throw SAMLComplianceException.create(SAMLCore_SamlExtensions,
            message = "An invalid SAML extension was found.",
            node = it)
```

library/src/main/kotlin/org/codice/compliance/Common.kt
brjeter: I like this much better. Putting this fix in my next PR 👍

library/src/main/kotlin/org/codice/compliance/Common.kt
```diff
@@ -144,6 +144,16 @@ fun Node.allChildren(name: String): List<Node> {
     return nodes
 }
 
+fun Node.attributeList(): List<Node> {
+    val attributesList = mutableListOf<Node>()
```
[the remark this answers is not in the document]
brjeter: Adding this to my next PR as well.

PR #97, opened 2018-04-15
Removed duplicate verification after schema verification (2/3)

blen-desta:

After the addition of https://github.com/connexta/saml-conformance/pull/90 some checks were being duplicated in code. This PR removes those. It also:
- removes unsupported protocol verifiers (ex. `NameIdManagementProtocolVerifier`)
- sperates the AuthnRequest specific verification to `AuthnRequestProtocolResponseVerifier`
- makes`LogoutRequestProtocolRequestVerifier` an extension `RequestProtocolVerifier`.

This is the second of 3 PRs
This PR covers sections 3 of the Core document
#### Checklist:


ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CommonDataTypeVerifier.kt
```diff
@@ -60,7 +60,7 @@ class CommonDataTypeVerifier {
 
         /** 1.3.1 String Values **/
         @Suppress("StringLiteralDuplication")
-        fun verifyStringValues(node: Node, errorCode: SAMLSpecRefMessage?) {
+        fun verifyStringValues(node: Node, errorCode: SAMLSpecRefMessage? = null) {
```
blen-desta: Changes for `CommonDataTypeVerifier` will be https://github.com/connexta/saml-conformance/pull/96

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/responses/AuthnRequestProtocolResponseVerifier.kt
```diff
@@ -0,0 +1,38 @@
...
+                                           override val acsUrl: String?) :
+        ResponseProtocolVerifier(response, id, acsUrl) {
+
+    /** 3.4 Authentication Request Protocol **/
+    override fun verify() {
+        verifyAuthnRequestProtocolResponse()
+    }
+
+    private fun verifyAuthnRequestProtocolResponse() {
+        if (response.children("Assertion").any { it.children("AuthnStatement").isEmpty() })
```
coyotesqrl: ❗️ This test looks wrong to me. This is checking if any assertion is missing an AuthnStatement; I believe it should be checking the inverse, that at least one assertion has an AuthnStatement
Maybe something like...
`if (response.children("Assertion").none { it.children("AuthnStatement").isNotEmpty() })`
That might work.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/ResponseProtocolVerifier.kt
```diff
@@ -15,155 +15,102 @@ package org.codice.compliance.verification.core
...
 
-        if (response.children(STATUS).isEmpty())
-            throw SAMLComplianceException.createWithXmlPropertyReqMessage(SAMLCore_3_2_2,
-                    property = STATUS,
-                    parent = RESPONSE,
-                    node = response)
+            CommonDataTypeVerifier.verifyUriValues(destination)
+        }
+
+        val content = response.attributes?.getNamedItem(DESTINATION)
```
SmithJosh: ❗️ This should be `consent`. And it's getting the destination, not consent

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/ResponseProtocolVerifier.kt
```diff
@@ -15,155 +15,102 @@ package org.codice.compliance.verification.core
...
+        }
     }
 
-    /**
-     * Verify the Statuses and Status Codes
-     * 3.2.2.1 Element <Status>
-     * 3.2.2.2 Element <StatusCode>
-     */
+    /** 3.2.2.2 Element <StatusCode> */
     private fun verifyStatusesType() {
```
SmithJosh: ✏️ Can we change this to verifyStatusType?

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/ResponseProtocolVerifier.kt
```diff
@@ -15,155 +15,102 @@ package org.codice.compliance.verification.core
...
-            if (statusCodes.any { it.attributes.getNamedItem(VALUE) == null })
-                throw SAMLComplianceException.createWithXmlPropertyReqMessage("SAMLCore.3.2.2.2",
-                        property = VALUE,
-                        parent = STATUS_CODE,
-                        node = response)
+        if (response.allChildren("Status")
+                        .any {
+                            !TOP_LEVEL_STATUS_CODES
+                                    .contains(it.children(STATUS_CODE).first()
+                                            .attributes.getNamedItem("Value").textContent)
```
SmithJosh: ❗️ The StatusCode is optional, so we may get a NoSuchElementException calling first blindly

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/responses/AuthnRequestProtocolResponseVerifier.kt
blen-desta: Ah, true

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/ResponseProtocolVerifier.kt
blen-desta: `<StatusCode>` is a required element for `<Status>`s (Core 3.2.2.1)

SmithJosh: Ah, sorry. Was looking at `3.2.2.2`

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/ResponseProtocolVerifier.kt
```diff
@@ -15,155 +15,102 @@ package org.codice.compliance.verification.core
...
-                    node = response)
-        CommonDataTypeVerifier.verifyIdValues(response.attributes.getNamedItem(ID),
+        CommonDataTypeVerifier.verifyIdValues(response.attributes.getNamedItem("ID"),
                 SAMLCore_3_2_2_a)
 
         // Assuming response is generated in response to a request
         val inResponseTo = response.attributes?.getNamedItem("InResponseTo")?.textContent
-        if (inResponseTo != id)
+        if (inResponseTo != null && inResponseTo != id)
             throw SAMLComplianceException.createWithPropertyMessage(SAMLCore_3_2_2_b,
```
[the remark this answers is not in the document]
blen-desta: I agree with adding a negative test. Are you saying we shouldn't test this then? Cause this will cover the:
```
Otherwise, it MUST be present and its value MUST match the value of the corresponding request's ID attribute.
```

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/responses/AuthnRequestProtocolResponseVerifier.kt
```diff
@@ -0,0 +1,39 @@
...
+        ResponseProtocolVerifier(response, id, acsUrl) {
+
+    /** 3.4 Authentication Request Protocol **/
+    override fun verify() {
+        verifyAuthnRequestProtocolResponse()
+    }
+
+    private fun verifyAuthnRequestProtocolResponse() {
+        if (response.children("Assertion")
+                        .none { it.children("AuthnStatement").isNotEmpty() })
```
SmithJosh: ✏️ Can we replace the double negative with all -> isEmpty?

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/RequestProtocolVerifier.kt
```diff
@@ -17,24 +17,17 @@ import org.codice.compliance.SAMLComplianceException
 import org.codice.compliance.SAMLCore_3_2_1_a
 import org.codice.compliance.SAMLCore_3_2_1_b
 import org.codice.compliance.SAMLCore_3_2_1_c
+import org.codice.compliance.utils.TestCommon.Companion.SAML_VERSION
 import org.codice.compliance.children
 import org.w3c.dom.Node
 
 abstract class RequestProtocolVerifier(open val request: Node) {
```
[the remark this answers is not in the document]
blen-desta: Removed

blen-desta:
build now

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/ResponseProtocolVerifier.kt
```diff
@@ -15,160 +15,105 @@ package org.codice.compliance.verification.core
...
-     */
+    /** All SAML responses are of types that are derived from the StatusResponseType complex type.*/
     private fun verifyStatusResponseType() {
-        if (response.attributes.getNamedItem(ID) == null)
-            throw SAMLComplianceException.createWithXmlPropertyReqMessage(SAMLCore_3_2_2,
-                    property = ID,
-                    parent = RESPONSE,
-                    node = response)
-        CommonDataTypeVerifier.verifyIdValues(response.attributes.getNamedItem(ID),
+        CommonDataTypeVerifier.verifyIdValues(response.attributes.getNamedItem("ID"),
```
coyotesqrl: ✏️ `response.attributes` is accessed here without safe nav. So either it is safe (and the safe accesses on lines 56, 76, and 91 don't need to be safe) or it will throw an NPE and it and the accesses on lines 64 and 74 need to be safe as well.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/ResponseProtocolVerifier.kt
```diff
@@ -15,160 +15,105 @@ package org.codice.compliance.verification.core
...
-        val destination = response.attributes?.getNamedItem("Destination")?.textContent
-        if (destination != acsUrl)
-            throw SAMLComplianceException.createWithPropertyMessage(SAMLCore_3_2_2_e,
-                    property = "Destination",
-                    actual = destination,
-                    expected = acsUrl ?: "No ACS URL Found",
-                    node = response)
+                response.attributes.getNamedItem("IssueInstant"), SAMLCore_3_2_2_d)
+
+        response.attributes?.getNamedItem(DESTINATION)?.let {
```
coyotesqrl: ✏️ A `let` is fine here, but an `apply` might be a little cleaner.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/ResponseProtocolVerifier.kt
```diff
@@ -15,160 +15,105 @@ package org.codice.compliance.verification.core
...
-                throw SAMLComplianceException.createWithXmlPropertyReqMessage("SAMLCore.3.2.2.2",
-                        property = VALUE,
-                        parent = STATUS_CODE,
-                        node = response)
+    /** 3.2.2.2 Element <StatusCode> */
+    private fun verifyStatusType() {
+        if (response.recursiveChildren("Status")
+                        .any {
+                            !TOP_LEVEL_STATUS_CODES
+                                    .contains(it.children(STATUS_CODE).first()
```
coyotesqrl: ❓ Do we need to protect against the case where there are no status code nodes?

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/responses/AuthnRequestProtocolResponseVerifier.kt
```diff
@@ -0,0 +1,39 @@
...
+import org.codice.compliance.SAMLComplianceException
+import org.codice.compliance.SAMLCore_3_4
+import org.codice.compliance.children
+import org.codice.compliance.verification.core.ResponseProtocolVerifier
+import org.w3c.dom.Node
+
+class AuthnRequestProtocolResponseVerifier(override val response: Node,
+                                           override val id: String,
+                                           override val acsUrl: String?) :
+        ResponseProtocolVerifier(response, id, acsUrl) {
```
coyotesqrl: ❓ I wonder if the base class should be renamed to `ProtocolResponseVerifier` or even just `ResponseVerifier` so the structure of its name matches its concrete subclasses.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/responses/AuthnRequestProtocolResponseVerifier.kt
```diff
@@ -0,0 +1,39 @@
...
+
+    /** 3.4 Authentication Request Protocol **/
+    override fun verify() {
+        verifyAuthnRequestProtocolResponse()
+    }
+
+    private fun verifyAuthnRequestProtocolResponse() {
+        if (response.children("Assertion")
```
coyotesqrl: ✏️ It reads a little oddly to me to test `all` `isEmpty`. Not that `none` `isNotEmpty` is that much better, but the use of none matches up with the "not found" a bit more, I think.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/ResponseProtocolVerifier.kt
blen-desta: `StatusCode` is a required element of `Status`

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/responses/AuthnRequestProtocolResponseVerifier.kt
blen-desta: `ResponseVerifier` it is

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/responses/AuthnRequestProtocolResponseVerifier.kt
blen-desta: I change it to `all` `isEmpty` because Josh commented that `none` `isNotEmpty` is a double negative and a bit confusing.

blen-desta:
build now

PR #98, opened 2018-04-16
Removed duplicate verification after schema verification (3/3)

blen-desta:

After the addition of https://github.com/connexta/saml-conformance/pull/90 some checks were being duplicated in code. This PR removes those. It also:

- Separated Signature and Encryption verification from CoreVerifier
- Removed unused saml exceptions in these sections

This is the last of 3 PRs
This PR covers the remaining sections of the Core document

#### Checklist:


ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SignatureAndEncryptionVerifier.kt
```diff
@@ -0,0 +1,86 @@
...
+                        message = "A signature needs to have exactly one Reference, " +
+                                "${references.size} found.",
+                        node = node)
+
+            val uriValue = references[0].attributes.getNamedItem("URI")?.textContent
+                    ?: throw SAMLComplianceException.create(SAMLCore_5_4_2_b1,
+                            message = "URI attribute not found.",
+                            node = node)
+
+            val formattedId = "#" + it.parentNode.attributes?.getNamedItem("ID")?.textContent
```
coyotesqrl: ✏️ String interpolation.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SignatureAndEncryptionVerifier.kt
```diff
@@ -0,0 +1,86 @@
...
+                        property = "URI",
+                        actual = uriValue,
+                        expected = formattedId,
+                        node = node)
+        }
+    }
+
+    /** 6.1 - General Considerations **/
+    private fun verifyGeneralConsiderations() {
+        val elements = mutableListOf<Node>()
```
coyotesqrl: ✏️ Instead of a mutable list with three calls to `addAll`, just initialize inline:
```kotlin
        val elements = node.children("EncryptedAssertion") + 
                node.children("EncryptedID") + 
                node.children("EncryptedAttribute")
```

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SignatureAndEncryptionVerifier.kt
```diff
@@ -0,0 +1,86 @@
...
+        elements.addAll(node.children("EncryptedAssertion"))
+        elements.addAll(node.children("EncryptedID"))
+        elements.addAll(node.children("EncryptedAttribute"))
```
SmithJosh: ❓ Don't we need to check all descendants here, not just children?

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SignatureAndEncryptionVerifier.kt
```diff
@@ -0,0 +1,85 @@
...
+        elements.flatMap { it.children("EncryptedData") }
+                .forEach {
+                    val type = it.attributes.getNamedItem("Type").textContent
+                    if (type != TestCommon.ELEMENT)
+                        throw SAMLComplianceException.createWithPropertyMessage(SAMLCore_6_1_b,
+                                property = "$ENCRYPTED_DATA's Type attribute",
+                                actual = type,
+                                expected = TestCommon.ELEMENT,
+                                node = node)
+                }
```
brjeter: ❗️ This whole function is already being covered by @bakejeyner's decrypt PR. Could you just add the `SAMLCore_6_1_b` message onto the existing messages in the Exception he throws in there? https://github.com/connexta/saml-conformance/blob/master/ctk/common/src/main/kotlin/org/codice/compliance/verification/core/EncryptionVerifier.kt#L70

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SignatureAndEncryptionVerifier.kt
```diff
@@ -0,0 +1,85 @@
...
+                        message = "A signature needs to have exactly one Reference, " +
+                                "${references.size} found.",
+                        node = node)
+
+            val uriValue = references[0].attributes.getNamedItem("URI")?.textContent
+                    ?: throw SAMLComplianceException.create(SAMLCore_5_4_2_b1,
+                            message = "URI attribute not found.",
+                            node = node)
+
+            val formattedId = "#${it.parentNode.attributes?.getNamedItem("ID")?.textContent}"
```
brjeter: ❗️ This assumes the `parentNode` of the Signature element is the root node, but that won't always be the case.  Should be `val formattedId = "#${node.attributes?.getNamedItem("ID")?.textContent}"` instead.

blen-desta: Not really. We want to get the direct parent of the Signature. A signature could be on the root response or inside an assertion. When it's inside the assertion we want to compare it with the assertion's ID not the root's. They are not the same values.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SignatureSyntaxAndProcessingVerifier.kt
```diff
@@ -0,0 +1,55 @@
...
+    private fun verifySignatureSyntaxAndProcessing() {
+        node.children(SSOConstants.SIGNATURE).forEach {
+            val references = it.recursiveChildren("Reference")
+            if (references.size != 1)
+                throw SAMLComplianceException.create(SAMLCore_5_4_2_b1,
+                        message = "A signature needs to have exactly one Reference, " +
+                                "${references.size} found.",
+                        node = node)
+
+            val uriValue = references[0].attributes.getNamedItem("URI")?.textContent
```
bakejeyner: ✏️ Instead of `references[0]`, could use `references.first()`.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SignatureAndEncryptionVerifier.kt
brjeter: You're right. I somehow overlooked "assertion" in the spec snippet. `...same-document reference to the ID attribute value of the root element of the assertion or protocol message being signed.`

blen-desta:
build now

blen-desta:
Aborted build
build now

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SignatureSyntaxAndProcessingVerifier.kt
```diff
@@ -0,0 +1,55 @@
...
+class SignatureSyntaxAndProcessingVerifier(private val node: Node) {
+
+    /** 5 SAML and XML Signature Syntax and Processing */
+    fun verify() {
+        verifySignatureSyntaxAndProcessing()
+    }
+
+    /** 5.4.2 References */
+    private fun verifySignatureSyntaxAndProcessing() {
+        node.children(SSOConstants.SIGNATURE).forEach {
```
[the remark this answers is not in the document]
blen-desta: I'll add that

PR #100, opened 2018-04-17
MAINT - Pull Node extension functions into their own class and adds new ones

coyotesqrl:
Adds two new extension functions for getting attribute nodes (one namespace-aware and one not) and two for getting attribute textContent (with and without namespace).

*I'm not sure if there is a lot of value in the extension functions for getting the attributes themselves and am open to discussion.*

#### Checklist:


coyotesqrl:
build now

PR #105, opened 2018-04-17
MAINT - Decrypting the SAML response in the tests

brjeter:
- Modified `verifyAndDecryptResponse` method to return Boolean
- Loop over the SAML message and run schema validation then decrypt
- Keycloak tests now pass! 🎉 
#### Checklist:

### DIFF
--- a/ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
+++ b/ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
@@ -24,6 +24,7 @@ import org.codice.compliance.attributeText
 import org.codice.compliance.children
 import org.codice.compliance.debugWithSupplier
 import org.codice.compliance.prettyPrintXml
+import org.codice.compliance.recursiveChildren
 import org.codice.compliance.utils.TestCommon
 import org.codice.compliance.utils.TestCommon.Companion.REQUESTER
 import org.codice.compliance.utils.schema.SchemaValidator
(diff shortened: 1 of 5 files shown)


ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
```diff
@@ -100,13 +103,26 @@ class CoreVerifier(val node: Node) {
...
+        fun decryptAndValidateSchema(responseDom: Node) {
+            val encVerifier = EncryptionVerifier()
+
+            do {
+                Log.debugWithSupplier {
+                    "Starting a pass of decryption and schema validation" +
+                            " on the SAML Response."
+                }
+                SchemaValidator.validateSAMLMessage(responseDom)
+            } while (encVerifier.verifyAndDecryptResponse(responseDom))
```
brjeter: Fixing to pass responseDom in EncryptionVerifier constructor instead.

brjeter:
@blen-desta I honestly wasn't sure where to put this method, and I like your idea way better. I'll go ahead and do that.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
coyotesqrl: ❓ Why is this in a loop? What is the exit condition for the loop?

brjeter: `verifyAndDecryptResponse` returns a Boolean on whether or not it found any encrypted elements. Because there may be encrypted elements within encrypted elements so we have to keep doing it until we find no more encrypted elements. That's super confusing though. Probably need to comment or javadoc that.

brjeter:
build now

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
```diff
@@ -104,25 +105,37 @@ class CoreVerifier(val node: Node) {
                         node = node)
         }
 
-        fun decryptAndValidateSchema(responseDom: Node) {
+        private fun preProcess(responseDom: Node) {
```
coyotesqrl: ✏️ While it's not very much better, I think this recursive version of this function is a little clearer.
```kotlin
        private fun preProcess(responseDom: Node,
                                encVerifier: EncryptionVerifier = EncryptionVerifier()) {
            val encElements = retrieveCurrentEncryptedElements(responseDom)
            if (encElements.isEmpty()) {
                Log.debugWithSupplier {
                    "Decrypted SAML Response:\n\n ${responseDom.prettyPrintXml()}"
                }
                return
            }

            Log.debugWithSupplier {
                "Starting a pass of decryption and schema validation" +
                        " on the SAML Response."
            }

            SchemaValidator.validateSAMLMessage(responseDom)
            encVerifier.verifyAndDecryptResponse(encElements)
            preProcess2(responseDom, encVerifier)
        }
```

brjeter: I like this better too. Fixed.

PR #106, opened 2018-04-18
CORE - 8.3.6 Validates entity identifier

coyotesqrl:
#### Checklist:
- [X] SAML Spec Table of Contents documentation updated

### DIFF
--- a/ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
+++ b/ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
@@ -17,6 +17,8 @@ import de.jupf.staticlog.Log
 import org.codice.compliance.SAMLComplianceException
 import org.codice.compliance.SAMLCoreRefMessage
 import org.codice.compliance.SAMLCore_3_2_1_d
+import org.codice.compliance.SAMLCore_8_3_6_a
+import org.codice.compliance.SAMLCore_8_3_6_b
 import org.codice.compliance.SAMLCore_SamlExtensions
 import org.codice.compliance.SAMLSpecRefMessage
 import org.codice.compliance.attributeNode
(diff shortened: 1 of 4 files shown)


ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
```diff
@@ -130,13 +134,37 @@ class CoreVerifier(val node: Node) {
...
+
+        private fun checkEntityIdentifier(node: Node) {
+            if (node.attributeNode("NameQualifier") != null ||
+                    node.attributeNode("SPNameQualifier") != null ||
+                    node.attributeNode("SPProvidedID") != null) {
+                throw SAMLComplianceException.create(SAMLCore_8_3_6_a,
+                        message = "No Subject element found.",
+                        node = node)
+            }
+            val nodeValue = node.nodeValue
```
bakejeyner: ✏️ Alternatively:
```
node.nodeValue?.let {
    if (it.length > ENTITY_ID_MAX_LEN)
        throw SAMLComplianceException.create(SAMLCore_8_3_6_b,
            message = "Length of URI [$it] is [${it.value}]",
            node = node)
}
```

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
```diff
@@ -130,13 +134,37 @@ class CoreVerifier(val node: Node) {
...
+            }
+        }
     }
 
     /**
      * Verify response against the Core Spec document
      */
     fun verify() {
         preProcess(node)
+        validateEntityIdentifiers(node)
```
bakejeyner: ✏️ `verifyEntityIdentifiers` instead of `validateEntityIdentifiers`. It might also make more sense to run `verifyCommonDataType` before the new method.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
coyotesqrl: Should have seen that when I pulled my conditions apart and extracted the `nodeValue` but I completely missed it. Nice.

coyotesqrl:
Second commit addresses review comments.

coyotesqrl:
build now

PR #107, opened 2018-04-19
CORE - 3.4.1.1 - Element <NameIDPolicy>

brjeter:
- Redesign `CoreVerifier`, `ResponseVerifier`, and `AuthnRequestProtocolResponseVerifier` to have `CoreVerifier` and `ResponseVerifier` be abstract. `AuthnRequestProtocolResponseVerifier` (renamed) extends `ResponseVerifier`
- Renamed `AuthnRequestProtocolResponseVerifier` to `CoreAuthnRequestProtocolVerifier`
- Reworked tests to allow me to put a nameIdPolicy in
- Added pluggable method in the `preProcess` method that can verify encrypted elements
- Added `NameIdPolicyVerifier`
#### Checklist:
- [X] SAML Spec Table of Contents documentation updated


ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
```diff
@@ -29,66 +28,49 @@ import org.codice.compliance.prettyPrintXml
...
+                throw SAMLComplianceException.createWithPropertyMessage(samlErrorCode,
+                        property = "Status",
+                        actual = "[]",
+                        expected = "<Status>",
+                        node = node)
+            status[0].children("StatusCode").forEach {
+                val code = it.attributeText("Value")
+
+                if (code != expectedStatusCode) {
+                    var exceptions = arrayOf(samlErrorCode)
```
coyotesqrl: ✏️ Alternately, use a conditional expression
```kotlin
                    val exceptions =
                            if (expectedStatusCode == TestCommon.REQUESTER)
                                arrayOf(samlErrorCode, SAMLCore_3_2_1_d)
                            else
                                arrayOf(samlErrorCode)
```

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/NameIDPolicyVerifier.kt
```diff
@@ -0,0 +1,77 @@
...
+import org.codice.compliance.SAMLCore_3_4_1_1c
+import org.codice.compliance.attributeText
+import org.codice.compliance.children
+import org.codice.compliance.recursiveChildren
+import org.codice.compliance.utils.TestCommon
+import org.codice.compliance.utils.TestCommon.Companion.ASSERTION
+import org.opensaml.saml.saml2.core.NameIDPolicy
+import org.w3c.dom.Node
+
+class NameIDPolicyVerifier(val response: Node, val policy: NameIDPolicy) {
```
coyotesqrl: ✏️ `policy` can be private.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/NameIDPolicyVerifier.kt
```diff
@@ -0,0 +1,77 @@
...
+                .forEach {
+                    val idFormat = it.attributeText("Format")
+                    if (!idFormat.equals(policy.format)) {
+                        throw SAMLComplianceException.create(SAMLCore_3_4_1_1b,
+                                message = "A NameID element was found with a Format attribute " +
+                                        "value of $idFormat instead of ${policy.format}.",
+                                node = it)
+                    }
+
+                    val spnq = it.attributeText("SPNameQualifier")
```
coyotesqrl: ✏️ Alternately (and I think a bit cleaner)
```kotlin
                    it.attributeText("SPNameQualifier")?.let {spnq -> 
                        if (spnq != policy.spNameQualifier) {
                            throw SAMLComplianceException.create(SAMLCore_3_4_1_1c,
                                    message = "A NameID element was found with a SPNameQualifier " +
                                            "attribute value of $spnq instead of " +
                                            "${policy.spNameQualifier}.",
                                    node = it)
                        }
                    }
```

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/NameIDPolicyVerifier.kt
```diff
@@ -0,0 +1,77 @@
...
+                                message = "A NameID element was found with a SPNameQualifier " +
+                                        "attribute value of $spnq instead of " +
+                                        "${policy.spNameQualifier}.",
+                                node = it)
+                    }
+                }
+    }
+
+    internal fun verifyEncryptedIds() {
+        if (policy?.format == TestCommon.NAMEID_ENCRYPTED) {
```
coyotesqrl: ✏️ Unnecessary safenav on `policy`.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlExtensionsVerifier.kt
```diff
@@ -0,0 +1,88 @@
...
+import org.codice.compliance.SAMLCore_SamlExtensions
+import org.codice.compliance.attributeList
+import org.codice.compliance.children
+import org.codice.compliance.recursiveChildren
+import org.codice.compliance.utils.TestCommon
+import org.w3c.dom.Attr
+import org.w3c.dom.Element
+import org.w3c.dom.Node
+
+class SamlExtensionsVerifier(val node: Node) {
```
coyotesqrl: ❓ Wouldn't schema validation be sufficient to affirm these extensions? At the least, it appears that `SubjectConfirmationData` is covered by the schema. Unless `lax` processing is laxer than I think, this snippet should ensure that no extended attributes or elements are added in the SAML namespace:
```xml
<element name="SubjectConfirmationData"
         type="saml:SubjectConfirmationDataType"/>
<complexType name="SubjectConfirmationDataType" mixed="true">
<complexContent>
    <restriction base="anyType">
        <sequence>
            <any namespace="##any" processContents="lax" minOccurs="0"
                 maxOccurs="unbounded"/>
        </sequence>
        <attribute name="NotBefore" type="dateTime" use="optional"/>
        <attribute name="NotOnOrAfter" type="dateTime" use="optional"/>
        <attribute name="Recipient" type="anyURI" use="optional"/>
        <attribute name="InResponseTo" type="NCName" use="optional"/>
        <attribute name="Address" type="string" use="optional"/>
        <anyAttribute namespace="##other" processContents="lax"/>
    </restriction>
</complexContent>
</complexType>
```

external/implementations/samlconf-keycloak-impl/README.md
```diff
@@ -4,6 +4,8 @@
 * Download Keycloak from [Keycloak Downloads](https://www.keycloak.org/downloads.html).
 * Unzip and run in standalone mode. Then setup an account with "admin" for both username and password. Then login.
 (Steps 2.1-2.4 in the [Keycloak Getting Started Guide](http://www.keycloak.org/docs/latest/getting_started/index.html#booting-the-server)).
+* Go to the Users tab and click the View All Users button.
+* Select the admin user and enter a valid email address under the Email field. Click Save.
```
coyotesqrl: ❓ Does it need to be valid, or does it just need to be in the form of an `addr-spec` as defined in [IETF RFC 2822](https://www.rfc-editor.org/info/rfc2822)? Either way, I think this instruction should be qualified to avoid users adding valid emails such as `John Doe <john.doe@mail.com>`, unless you've confirmed that Keycloak is correctly converting those values to the `addr-spec` format.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/NameIDPolicyVerifier.kt
```diff
@@ -0,0 +1,77 @@
...
+import org.codice.compliance.utils.TestCommon
+import org.codice.compliance.utils.TestCommon.Companion.ASSERTION
+import org.opensaml.saml.saml2.core.NameIDPolicy
+import org.w3c.dom.Node
+
+class NameIDPolicyVerifier(val response: Node, val policy: NameIDPolicy) {
+
+    /** 3.4.1.1 Element <NameIDPolicy> **/
+    internal fun verify() {
+        response
```
blen-desta: ❓ Should we add a check to verify that the format is not `urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified` or `urn:oasis:names:tc:SAML:2.0:nameid-format:encrypted`?

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/NameIDPolicyVerifier.kt
```diff
@@ -0,0 +1,77 @@
...
+                                        "attribute value of $spnq instead of " +
+                                        "${policy.spNameQualifier}.",
+                                node = it)
+                    }
+                }
+    }
+
+    internal fun verifyEncryptedIds() {
+        if (policy?.format == TestCommon.NAMEID_ENCRYPTED) {
+            val subjects = response.recursiveChildren(TestCommon.ASSERTION)
```
blen-desta: ✏️ I think you can just use `ASSERTION` here like above

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlExtensionsVerifier.kt
brjeter: I will check on that.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/NameIDPolicyVerifier.kt
brjeter: Yes, good catch.

PR #108, opened 2018-04-19
CORE - Adds JTidy to clean up HTML responses and make them valid.

coyotesqrl:
HTML responses should only be returned in unusual error conditions, so this change will confirm if the response is HTML and only then tidy it before parsing occurs.

#### Checklist:


coyotesqrl:
build now

PR #111, opened 2018-04-20
MANT - Subject Comparison Framework

bakejeyner:
#### Checklist:
- [ ] SAML Spec Table of Contents documentation updated **<- No tests added**

### DIFF
--- a/ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
+++ b/ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
@@ -1,0 +1,252 @@
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.compliance.verification.core
+
+import org.codice.compliance.SAMLComplianceException
+import org.codice.compliance.SAMLCore_3_3_4_b
+import org.codice.compliance.SAMLCore_3_3_4_c
+import org.codice.compliance.SAMLCore_3_4_1_4_b
+import org.codice.compliance.SAMLProfiles_4_1_4_2_e
+import org.codice.compliance.attributeList
+import org.codice.compliance.attributeText
+import org.codice.compliance.children
+import org.codice.compliance.recursiveChildren
+import org.codice.compliance.utils.TestCommon.Companion.FORMAT
+import org.codice.compliance.utils.TestCommon.Companion.SUBJECT
+import org.codice.compliance.utils.TestCommon.Companion.SUBJECT_CONFIRMATION
+import org.w3c.dom.Node
+
+/**
+ * Verifies the request subjects **strongly match** the response subject according to the Core
+ * document sections 3.3.4 & 3.4.1.4. (verifySubjectsMatchSSO)
+ *
+ * Verifies the response subjects refer to the same principal according to the Profiles document
+ * section 4.1.4.2. (verifySubjectsMatchAuthnRequest)
+ *
+ * @param request Optional AuthnRequest {@code Node}. Should be passed when calling the
+ * verifySubjectsMatchAuthnRequest method.
+ * @param response Response {@code Node}.
+ */
+class SubjectComparisonVerifier(private val request: Node? = null, private val response: Node) {
+
+    companion object {
+        private const val UNSPECIFIED_URI =
+                "urn:oasis:names:tc:SAML:2.0:nameid-format:unspecified"
+        private const val EMPTY_STRING = ""
+    }
+
+    /**
+     * This function verifies that all of the Subjects in the Response refer to the same principal.
+     *
+     * NOTE: This method does not actually resolve who the principal of the subject is. It only
+     * tests that if two Subject identifiers have the same format, their content is the same. In
+     * order to fully test this function we need to resolve the Subjects to a principal.
+     */
+    fun verifySubjectsMatchSSO() {
+        val subjectList = response.children("Assertion")
+                .flatMap { it.children(SUBJECT) }
+                .toList()
+
+        subjectList.forEachIndexed { index, outerSubject ->
+            subjectList.subList(index + 1, subjectList.size).forEach { innerSubject ->
+                verifyIdContentsMatchSSO(outerSubject, innerSubject)
+            }
+        }
+    }
+
+    /**
+     * Compares the text content of the identifiers. Special checking due to nameIdPolicyFormat.
+     */
+    private fun verifyIdContentsMatchSSO(id1: Node, id2: Node) {
+        val format1 = id1.filteredFormatValue()
+        val format2 = id2.filteredFormatValue()
+
+        // If they share a format value that isn't "unspecified", but have different contents
+        if (format1 == format2
+                && format1 != null
+                && id1.textContent != id2.textContent) {
+            throw SAMLComplianceException.create(SAMLProfiles_4_1_4_2_e,
+                    SAMLCore_3_3_4_b,
+                    message = "Two Response Subject identifiers have identical Format attributes " +
+                            "[$format1], but the content of one [${id1.textContent}] is not " +
+                            "equal to the content of the other [${id2.textContent}]",
+                    node = response)
+        }
+    }
+
+    /**
+     * This function verifies that the Subjects from the Response **strongly match** the Subject
+     * from the AuthnRequest.
+     *
+     * NOTE: This method does not test to make sure the Response Subject's identifier's Format is
+     * equal to an optional NameIDPolicyFormat from the AuthnRequest. That is tested in
+     * NameIDPolicyVerifier
+     *
+     * NOTE: This method does not actually resolve who the principal of the subject is. It only
+     * compares the given Subject values to each other. In order to fully test this function we
+     * need to resolve the Subjects to a principal.
+     */
+    @Suppress("NestedBlockDepth" /* Simple `let` nesting */)
+    fun verifySubjectsMatchAuthnRequest() {
+
+        val requestSubject = request?.children("Subject")?.firstOrNull() ?: return
+        val requestId = requestSubject.getId()
+        val requestConfirmations = requestSubject.children(SUBJECT_CONFIRMATION)
+
+        if (requestId == null && requestConfirmations.isEmpty()) return
+
+        val nameIdPolicyFormat =
+                request.children("NameIDPolicy").firstOrNull()?.filteredFormatValue()
+
+        response.recursiveChildren("Assertion")
+                .flatMap { it.children(SUBJECT) }
+                .forEach { resSubject ->
+
+                    // Verify ids match
+                    requestId?.let { reqId ->
+                        resSubject.getId()?.let { resId ->
+                            verifyIdAttributesMatchAuthnRequest(reqId, resId, nameIdPolicyFormat)
+                            verifyIdContentsMatchAuthnRequest(reqId, resId)
+                        } ?: throw SAMLComplianceException.create(
+                                SAMLCore_3_3_4_b,
+                                SAMLCore_3_4_1_4_b,
+                                message = "One of the Response's Subjects contained no identifier",
+                                node = resSubject)
+                    }
+
+                    // Verify SubjectConfirmations match
+                    requestConfirmations.let { reqConfirmations ->
+                        resSubject.children(SUBJECT_CONFIRMATION).let { resConfirmations ->
+                            if (resConfirmations.isEmpty())
+                                throw SAMLComplianceException.create(
+                                        SAMLCore_3_3_4_c,
+                                        SAMLCore_3_4_1_4_b,
+                                        message = "One of the Response's Subjects contained no " +
+                                                "SubjectConfirmations.",
+                                        node = resSubject)
+                            verifyConfirmationsMatchAuthnRequest(reqConfirmations,
+                                    resConfirmations)
+                        }
+                    }
+                }
+    }
+
+    /**
+     * Compares all of the attributes of the identifiers. Special checking for the Format attribute.
+     */
+    private fun verifyIdAttributesMatchAuthnRequest(reqId: Node,
+                                                    resId: Node,
+                                                    nameIdPolicyFormat: String? = null) {
+        // Check the format attribute separately
+        /*
+         * If no NameIDPolicyFormat is defined, the Request format and the Response format are
+         * different, and the formats are not unspecified
+         */
+        if (nameIdPolicyFormat == null) {
+            val reqFormat = reqId.filteredFormatValue()
+            val resFormat = resId.filteredFormatValue()
+            if (reqFormat != null && reqFormat != resFormat)
+                throw SAMLComplianceException.create(SAMLCore_3_3_4_b,
+                        message = "One of the Response's Subject identifier's Format attribute " +
+                                "[${resFormat ?: UNSPECIFIED_URI}] is not identical to the " +
+                                "AuthnRequest's Subject identifier's Format attribute " +
+                                "[$reqFormat].",
+                        node = resId)
+        }
+
+        val reqAttributes = reqId.attributeList().filter { it.localName != FORMAT }.toSet()
+        val resAttributes = resId.attributeList().filter { it.localName != FORMAT }.toSet()
+
+        if (reqAttributes != resAttributes)
+            throw SAMLComplianceException.create(SAMLCore_3_3_4_b,
+                    SAMLCore_3_4_1_4_b,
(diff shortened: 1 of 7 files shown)


bakejeyner:
**Everyone except @brjeter should only review for syntax, unless you want to learn this stuff**

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
```diff
@@ -0,0 +1,241 @@
...
+    /**
+     * Verifies that the request SubjectConfirmations **strongly match** the response confirmations.
+     *
+     * Note that this matching only compares the methods of the SubjectConfirmations. In order to
+     * properly strongly match we would need to "confirm" the SubjectConfirmation (Hard To Test)
+     */
+    private fun verifyConfirmationsMatchAuthnRequest(requestConfirmations: List<Node>,
+                                                     responseConfirmations: List<Node>) {
+
+        val requestMethods = requestConfirmations.map { it.attributeText("Method") }.toSet()
```
coyotesqrl: ❓ Should any potential nulls be filtered out here? Could there be a null?

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
```diff
@@ -0,0 +1,241 @@
...
+        val responseMethods = responseConfirmations.map { it.attributeText("Method") }.toSet()
+
+        // If none the Request SubjectConfirmations have any matching Response SubjectConfirmations
+        if (requestMethods.none { reqMethod ->
```
coyotesqrl: ❓ Wouldn't the following test be equivalent (and clearer)?
```kotlin
        if (requestMethods.none { reqMethod ->
                    responseMethods.contains(reqMethod)
                })
```

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
```diff
@@ -0,0 +1,241 @@
...
+            when (it) {
+                EMPTY_STRING -> null
+                UNSPECIFIED_URI -> null
+                else -> it
+            }
+        }
+    }
+
+    private fun Node.getId(): Node? {
+        return this.children().firstOrNull { it.localName != "SubjectConfirmation" }
```
coyotesqrl: ❓ I don't understand this function.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    private fun Node.getSubjectConfirmations(): List<Node>? {
```
coyotesqrl: ✏️ I recognize that you're using the `null` state of this value as a quick conditional check, but it's a bit more fragile than returning an empty list. As it's private, it's not a big deal; however, we should try to reduce the nulls we pass around, not increase them.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
```diff
@@ -0,0 +1,241 @@
...
+ * Tests if request **strongly matches** response according to the Core document sections 3.3.4 &
+ * 3.4.1.4
+ *
+ * @param request First subject to compare
+ * @param response Response
+ * @param authnRequest An optional AuthnRequest to use for its NameIDPolicy when comparing the
+ * subject of an AuthnRequest to the subject of a Response Assertion
+ * (see Core 3.4.1.4 for more information)
+ */
+class SubjectComparisonVerifier(private val request: Node? = null, private val response: Node) {
```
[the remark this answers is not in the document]
bakejeyner: Correct. This PR only adds the verification framework.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
```diff
@@ -0,0 +1,241 @@
...
+        val nameIdPolicyFormat =
+                request.children("NameIDPolicy").firstOrNull()?.getFilteredFormatValue()
+
+        response.recursiveChildren("Assertion")
+                .flatMap { it.children(SUBJECT) }
+                .forEach { resSubject ->
+
+                    // Verify ids match
+                    requestId?.let { reqId ->
+                        resSubject.getId()?.let { resId ->
```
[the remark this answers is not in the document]
bakejeyner: >Can the subject have a null id if the requestId is null?
Correct. The Response subject can have a null id if the request id is null. There are no restraints on the Response subject in such a case (other than that the subjects must all refer to the same principal, but that is tested in the SSO version of this method).
`If S2 includes an identifier element (<BaseID>, <NameID>, or <EncryptedID>), then S1 MUST include an identical identifier element...`
>Can the subject have a non-null id if the requestId is null?
Correct. The Response subject can have a non-null id if the request id is null (See above reference).

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
```diff
@@ -0,0 +1,241 @@
...
+        verifyIdContentsMatchAuthnRequest(requestId, responseId)
+    }
+
+    /**
+     * Compares all of the attributes of the identifiers. Special checking for the Format attribute.
+     */
+    private fun verifyIdAttributesMatchAuthnRequest(reqId: Node,
+                                                    resId: Node,
+                                                    nameIdPolicyFormat: String? = null) {
+        val reqFormat = reqId.getFilteredFormatValue()
```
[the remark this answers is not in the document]
bakejeyner: I'm a little confused by this comment. I'll contact you in person when I'm ready!

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
```diff
@@ -0,0 +1,241 @@
...
+        }
+    }
+
+    /**
+     * Verifies that the request SubjectConfirmations **strongly match** the response confirmations.
+     *
+     * Note that this matching only compares the methods of the SubjectConfirmations. In order to
+     * properly strongly match we would need to "confirm" the SubjectConfirmation (Hard To Test)
+     */
+    private fun verifyConfirmationsMatchAuthnRequest(requestConfirmations: List<Node>,
```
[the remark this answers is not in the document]
bakejeyner: I'll leave it as is, but that is a good observation.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
bakejeyner: I personally find more value in Kotlin passing around nulls. Nulls offer more alternatives than empty lists. Purely my opinion though.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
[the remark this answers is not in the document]
bakejeyner: Good to go!

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
bakejeyner: After discussion with Richard this is the conclusion we came to (summarized by @bakejeyner):
`Avoid null values when you can. Kotlin _does_ have fancy null stuff, but only because nulls are poorly handled in Java. Rather than using that syntax to make cool looking code, try not to use it because it is only there to fix the mistakes other languages make.`

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
```diff
@@ -0,0 +1,267 @@
...
+    /**
+     * This function verifies that all of the Subjects in the Response refer to the same principal.
+     *
+     * NOTE: This method does not actually resolve who the principal of the subject is. It only
+     * tests that if two Subject identifiers have the same format, their content is the same. In
+     * order to fully test this function we need to resolve the Subjects to a principal.
+     *
+     * NOTE: This method does not actually resolve who the principal of the subject is. It only
+     * compares the given Subject values to each other.
+     */
```
brjeter: ❗️ You have two redundant "NOTE"s here. If you decided to keep the first one, the second sentence has some weird wording.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
```diff
@@ -0,0 +1,267 @@
...
+    fun verifySubjectsMatchSSO() {
+        val subjectList = response.children("Assertion")
+                .flatMap { it.children(SUBJECT) }
+                .toList()
+
+        subjectList.forEachIndexed { index, outerSubject ->
+            subjectList.subList(index + 1, subjectList.size).forEach { innerSubject ->
```
brjeter: ❗️ I thought we decided you didn't need two nested loops here and that you could just do one pass.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    private fun verifyIdsMatchAuthnRequest(
+            requestId: Node,
+            responseId: Node,
+            nameIdPolicyFormat: String?) {
+
+        verifyIdAttributesMatchAuthnRequest(requestId, responseId, nameIdPolicyFormat)
+        verifyIdContentsMatchAuthnRequest(requestId, responseId)
+    }
```
brjeter: ✏️ This private function feels unnecessary to me. In its one usage above, I think it's just as clear to inline the call to the `verifyIdAttributesMatchAuthnRequest` and `verifyIdContentsMatchAuthnRequest` methods.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
```diff
@@ -0,0 +1,268 @@
...
+                        node = resId)
+        }
+
+        val reqAttributes = reqId.attributeList().filter { it.localName != FORMAT }.toSet()
+        val resAttributes = resId.attributeList().filter { it.localName != FORMAT }.toSet()
+
+        if (reqAttributes != resAttributes)
+            throw SAMLComplianceException.create(SAMLCore_3_3_4_b2,
+                    SAMLCore_3_4_1_4_b1,
+                    message = "One of the Response'a Subject identifier's attributes " +
```
brjeter: ✏️ "Response's" (there's an "a" instead of an "s")

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
```diff
@@ -0,0 +1,268 @@
...
+
+    /**
+     * Verifies that the request SubjectConfirmations **strongly match** the response confirmations.
+     *
+     * Note that this matching only compares the methods of the SubjectConfirmations. In order to
+     * properly strongly match we would need to "confirm" the SubjectConfirmation (Hard To Test)
+     */
+    private fun verifyConfirmationsMatchAuthnRequest(requestConfirmations: List<Node>,
+                                                     responseConfirmations: List<Node>) {
+
```
brjeter: ✏️ It may be a good idea to have require statements that verify that these are, in fact, `SubjectConfirmation` nodes. Just as a further check to keep us from shooting ourselves in the foot down the road (i.e. running this method on the wrong Nodes which would give us a false positive).

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
```diff
@@ -0,0 +1,265 @@
...
+        val requestMethods = requestConfirmations
+                .map { it.attributeText("Method") }
+                .filter { it != null }.toSet()
+        val responseMethods = responseConfirmations
+                .map { it.attributeText("Method") }
+                .filter { it != null }.toSet().toSet()
```
brjeter: ✏️ For this line and the usage above, you can just call `filterNotNull()` instead of `filter { it != null }`.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
```diff
@@ -0,0 +1,265 @@
...
+
+        // If none the Request SubjectConfirmations have any matching Response SubjectConfirmations
```
brjeter: ✏️ "If none of..."

library/src/main/resources/SAMLSpecRefMessage.properties
```diff
@@ -216,12 +249,23 @@ SAMLCore_3_4_1_1c=When a Format defined in Section 8.3 other than \
...
-SAMLCore_3_4_1_4b=At least one statement in at least one assertion MUST be a <saml:AuthnStatement> \
+
+SAMLCore_3_4_1_4_b=If the <saml:Subject> element in the request is present, then the resulting \
+  assertions' <saml:Subject> MUST strongly match the request <saml:Subject>, as described in \
+  Section 3.3.4, except that the identifier MAY be in a different format if specified by \
+  <NameIDPolicy>.
+
+SAMLCore_3_4_1_4_b1=If the <saml:Subject> element in the request is present, then the resulting \
+  assertions' <saml:Subject> MUST strongly match the request <saml:Subject>, as described in \
+  Section 3.3.4
```
brjeter: ❗️ I don't like breaking spec references up like this so that the messaging can be more specific. I don't like it for a few reasons:
1. A more targeted message is what the `message` property in the `SamlComplianceException` is for.
2. We end up with the overall messages in here which are not being used at all.
3. It adds more complexity to understanding the code as a developer.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
```diff
@@ -0,0 +1,265 @@
...
+     *
+     * NOTE: This method does not actually resolve who the principal of the subject is. It only
+     * tests that if two Subject identifiers have the same format, their content is the same. In
+     * order to fully test this function we need to resolve the Subjects to a principal.
+     *
+     * NOTE: This method does not actually resolve who the principal of the subject is. It only
+     * compares the given Subject values to each other.
+     */
+    fun verifySubjectsMatchSSO() {
+        val subjectList = response.children("Assertion")
```
coyotesqrl: ✏️ Using Guava, can simplify this fun to either
```kotlin
        Sets.combinations(response.children("Assertion")
                .flatMap { it.children(SUBJECT) }
                .toSet(), 2).forEach {
            verifyIdContentsMatchSSO(it.first(), it.last())
        }
```
or 
```kotlin
        val subjectList = response.children("Assertion")
                .flatMap { it.children(SUBJECT) }
                .toSet()

        Sets.combinations(subjectList, 2).forEach {
            verifyIdContentsMatchSSO(it.first(), it.last())
        }
```

It's a bit less efficient (for our small sets is marginally so), but gets us out of the business of doing the combinatoric work manually.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
bakejeyner: For now I'm going to leave the method as is, reasoning stated below.

I feel like it would be weird to have require statements like that in this private method if we don't have similar require statements in other functions in our codebase. This may be a good topic for discussion.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
coyotesqrl: ✏️ While a little less efficient, for the size of our sets this is a lot simpler and gets us out of the business of managing combinatorial math.
```kotlin
        val subjectSet = response.children("Assertion")
                .flatMap { it.children(SUBJECT) }
                .toSet()
        Sets.combinations(subjectSet, 2).forEach {
            verifyIdContentsMatchSSO(it.first(), it.last())
        }
```
Can also inline the `subjectSet` to make it even more concise.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
bakejeyner: Chose the second option because it looks more readable in that form.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
SmithJosh: ✏️ The verifyIdContentsMatchSSO check appears to be transitive, so I believe the following should accomplish the same thing and avoid checking all combinations. 
```
subjectList.forEachIndexed { i, outerSubject ->
            val innerSubject = subjectList.get(i + 1)
            verifyIdContentsMatchSSO(outerSubject, innerSubject)
}
```

bakejeyner:
build now

PR #112, opened 2018-04-23
MAINT - Pulling setup logic into TestCommon for reuse

brjeter:
#### Checklist:


ctk/common/src/main/kotlin/org/codice/compliance/utils/TestCommon.kt
```diff
@@ -146,6 +154,77 @@ class TestCommon {
...
             }.build()
         }
+
+        /**
+         * Provides a default request for testing
+         * @return A valid Redirect AuthnRequest.
+         */
+        fun createDefaultAuthnRequest(binding: SamlProtocol.Binding): AuthnRequest {
+            return AuthnRequestBuilder().buildObject().apply {
+                issuer = IssuerBuilder().buildObject().apply {
```
bakejeyner: ✏️ If it fits on one line, it could look cleaner as:
```
issuer = IssuerBuilder().buildObject().apply { value = SP_ISSUER }

PR #116, opened 2018-04-25
MAINT - Upgrades to JUnit 5 and Gradle 4.7

coyotesqrl:
Upgrades kotlintest to latest version.
Removes suites module and updates Command.kt to work directly with classes.
Adds one test class for common verifier.

##### Heroing Steps
This PR will need to be heroed, the more the better.

* Update new test class to put in a failure condition and confirm that it fails
* Update code under test for new test class, putting in a failure condition and confirm the test fails
* Update one of the extant library junit test classes to add a failure condition
* Update code under test for one of the extant library junit test classes
* Update one of the CTK test classes with a failure condition and confirm it fails
* Update a verifier beneath a CTK test class with a failure condition
* Confirm that the Error test classes are invoked when the `-e` flag is passed to the command line
* Verify that the Docker tests against DDF pass

A truly energized hero would also:
* See if the new test class - `BindingVerifierSpec` - reports three passed tests as it was for me; if so, find the cause of the `foo()` test being reported when it does not appear to exist
* Test cleaning up the main gradle file, paying particular regard to the `testImplementation` and `testRuntimeOnly` function calls within `subprojects/dependencies`. This is a delicate process, and one must ensure that each of the following all still run and report correctly:
    * The new test class
    * The extant test classes in `library`
    * The CTK tests

#### Checklist:


build.gradle
```diff
@@ -18,20 +18,23 @@ plugins {
...
-    kotlin_test_version = '2.0.7'
+    kotlin_test_version = '3.1.0-RC2'
     rest_assured_version = '2.9.0'
     org_slf4j_version = '1.7.1'
-    junit_version = '4.12'
     guava_version = '24.0-jre'
     detekt_version = '1.0.0.RC6-4'
     java_encoding = 'UTF-8'
     static_log_version = '2.2.0'
+    mockk_version = '1.7.15'
```
brjeter: ✏️ It doesn't look like this is being used anywhere

build.gradle
```diff
@@ -62,13 +66,26 @@ subprojects {
...
+        compile "io.kotlintest:kotlintest-runner-junit5:$kotlin_test_version"
+        compile "org.junit.platform:junit-platform-suite-api:$junit_platform_version"
+        compile "org.junit.platform:junit-platform-runner:$junit_platform_version"
+
+        testImplementation(
+                "org.junit.jupiter:junit-jupiter-api:$junit_jupiter_version"
+        )
+        testRuntimeOnly(
+                "org.junit.jupiter:junit-jupiter-engine:$junit_jupiter_version",
+        )
```
brjeter: ✏️ The gradle release notes on JUnit5 support has it like
`testImplementation 'org.junit.jupiter:junit-jupiter-api:5.1.0'` (instead of having the parentheses).
✏️ Also you have an unneeded comma at the end of line 83.

deployment/distribution/build.gradle
```diff
@@ -15,7 +15,6 @@ dependencies {
     compile project(':external:api')
     compile project(':ctk:common')
     compile project(':ctk:idp')
-    compile project(':deployment:suites')
```
brjeter: ✏️ There's a section for the suites directory in the README that should be removed.

deployment/distribution/src/main/kotlin/org/codice/ctk/Command.kt
```diff
@@ -56,7 +74,24 @@ fun main(args: Array<String>) {
...
+
+    val summaryGeneratingListener = SummaryGeneratingListener()
+    LauncherFactory.create().apply {
+        registerTestExecutionListeners(summaryGeneratingListener)
+    }.execute(request)
+
+    PrintWriter(System.out).use {
+        summaryGeneratingListener.summary.printFailuresTo(it)
+        summaryGeneratingListener.summary.printTo(it)
+    }
```
brjeter: ❓ Is this adding functionality or just replicating the functionality we used to have?

ctk/common/src/test/kotlin/org/codice/compilance/verification/binding/BindingVerifierSpec.kt
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        "Check Error Status Code is not Error" {
+            forAll(codeSeq) { code ->
+                if (code >= HttpURLConnection.HTTP_BAD_REQUEST) {
+                    shouldThrow<SAMLComplianceException> {
+                        BindingVerifier.verifyHttpStatusCodeErrorResponse(code)
+                    }
+                } else {
+                    code shouldBe code
```
brjeter: ✏️ Shouldn't this line be still calling the `BindingVerifier` method and just making sure no exceptions are thrown? This comment applies to both tests.

build.gradle
```diff
@@ -18,20 +18,23 @@ plugins {
     id 'io.gitlab.arturbosch.detekt' version '1.0.0.RC6-4'
     id 'com.diffplug.gradle.spotless' version '3.10.0'
     id 'net.ltgt.errorprone' version '0.0.13'
+    id "com.adarshr.test-logger" version "1.2.0"
```
brjeter: After looking into what this is, I'm now excited 😄

build.gradle
coyotesqrl: It will likely be in use when I write my next test. If not, I'll remove it then.

deployment/distribution/src/main/kotlin/org/codice/ctk/Command.kt
coyotesqrl: It is the new way of generating the output.

ctk/common/src/test/kotlin/org/codice/compilance/verification/binding/BindingVerifierSpec.kt
coyotesqrl: I don't know how to write this half of the tests. I _definitely should_ be calling the verifier methods, so thanks for reminding me, but there's no `notThrown` method so I'll want to put something in to indicate success.

build.gradle
coyotesqrl: I didn't configure it at all, but even out of the box I think it looks better.

build.gradle
coyotesqrl: Will check and will fix.

coyotesqrl:
Commit pushed with code review changes.

blen-desta:
build now

PR #117, opened 2018-04-25
[WIP] CORE 3.2.2 - Added signature verification for POST response

blen-desta:
Validating the POST response's signature.

This is a WIP PR. I am creating it to get your opinions about using objects instead of `Node`. DDF's `SimpleSign` takes in a `Signature` object so this may be a good way of validating the signature. Moving forward we can gradually convert passing `Node`s to actual object too.

#### Checklist:

### DIFF
--- a/library/src/main/java/org/codice/security/sign/SimpleSign.java
+++ b/library/src/main/java/org/codice/security/sign/SimpleSign.java
@@ -34,15 +34,26 @@ import org.apache.cxf.rs.security.saml.sso.SSOConstants;
 import org.apache.wss4j.common.crypto.CryptoType;
 import org.apache.wss4j.common.ext.WSSecurityException;
 import org.apache.wss4j.common.saml.OpenSAMLUtil;
+import org.apache.wss4j.common.saml.SAMLKeyInfo;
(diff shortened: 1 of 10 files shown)


library/src/main/java/org/codice/security/sign/SimpleSign.java
```diff
@@ -241,6 +252,76 @@ public boolean validateSignature(
...
+      SignatureValidator.validate(signature, credential);
+    } catch (org.opensaml.xmlsec.signature.support.SignatureException e) {
+      throw new SignatureException("Error validating the XML signature", e);
+    } finally {
+      if (threadLoader != null) {
+        Thread.currentThread().setContextClassLoader(threadLoader);
+      }
+    }
+  }
+
```
brjeter: ❓ Did all of this code come from DDF?

blen-desta: It sure did

blen-desta:
I am closing this PR because the changes @brjeter is working on will impact it. Will put another one up once his changes are merged.

PR #126, opened 2018-05-02
CORE - 3.4.1 - Element <AuthnRequest>

brjeter:
- Added method to check the ACS url (using the EntityInformation object to do the work)
- Changed Core and Profiles verifiers to take AuthnRequest objects instead of request dom. This simplified the retrieval of the ACS. Other option was to turn the dom back into an AuthnRequest object in the Core verifier.
- Copy-pasted the new EntityInformation class from DDF (since Blen updated it)
- Reorganized the tests to read consistently
#### Checklist:


ctk/common/src/main/kotlin/org/codice/compliance/verification/binding/PostFormVerifier.kt
```diff
@@ -175,8 +173,7 @@ class PostFormVerifier(private val response: Response, private val isRelayStateG
     @Suppress("ComplexMethod", "NestedBlockDepth")
     private fun verifyPostForm() {
         responseForm?.let {
-            if (it.hasNoAttributeWithNameAndValue(ACTION,
-                            checkNotNull(TestCommon.acsUrl[HTTP_POST]))) {
+            if (it.getAttribute(ACTION).isNullOrEmpty()) {
```
bakejeyner: ❓ Why was the comparison to the`acsUrl` taken out?

library/src/main/java/org/codice/security/saml/EntityInformation.java
```diff
@@ -94,45 +90,38 @@ Binding getBinding(AuthnRequest request, Binding preferred) {
...
-            Collectors.toMap(AbstractMap.SimpleEntry::getKey, AbstractMap.SimpleEntry::getValue));
-  }
-
   public ServiceInfo getAssertionConsumerService(Binding binding) {
     return assertionConsumerServices.get(binding);
   }
 
   public ServiceInfo getAssertionConsumerService(
       AuthnRequest request, Binding preferred, Integer index) {
     ServiceInfo si = null;
```
bakejeyner: ✏️ Super nitpick: we don't use acronym variable names a lot in the codebase. Probably more consistent to change `si` to `serviceInfo` or something of that manner instead.

ctk/common/src/main/kotlin/org/codice/compliance/utils/RestAssuredExtensions.kt
```diff
@@ -25,20 +26,38 @@ import org.codice.compliance.verification.binding.PostBindingVerifier
...
+    return when (this.determineBinding()) {
+        SamlProtocol.Binding.HTTP_REDIRECT -> {
+            // Extracted according to Binding 3.4.4
+            this.getHeader("Location")
+        }
+        SamlProtocol.Binding.HTTP_POST -> {
+            // Extracted according to Bindings 3.5.4
+            this.extractSamlResponseForm()?.getAttribute(ACTION)
+        }
+        else -> throw UnsupportedOperationException(BINDING_UNSUPPORTED_MESSAGE)
```
coyotesqrl: ❓ I wonder if it might be clearer/better if we added a specific case for artifact binding even though it would just do the same as the else branch.

library/src/main/java/org/codice/security/saml/EntityInformation.java
brjeter: I think I'd rather leave this as it is in DDF.

ctk/common/src/main/kotlin/org/codice/compliance/verification/binding/PostFormVerifier.kt
brjeter: Because I'm effectively checking it now in the `CoreAuthnRequest` verifier. It's that weirdness where the check applies to both Core and Bindings.

ctk/common/src/main/kotlin/org/codice/compliance/utils/RestAssuredExtensions.kt
```diff
@@ -25,20 +26,38 @@ import org.codice.compliance.verification.binding.PostBindingVerifier
...
 fun Response.getBindingVerifier(): BindingVerifier {
     return when (this.determineBinding()) {
         SamlProtocol.Binding.HTTP_REDIRECT -> RedirectBindingVerifier(this)
         SamlProtocol.Binding.HTTP_POST -> PostBindingVerifier(this)
-        else -> throw UnsupportedOperationException("Binding is not currently supported.")
+        else -> throw UnsupportedOperationException(BINDING_UNSUPPORTED_MESSAGE)
+    }
+}
+
+fun Response.getLocation(): String? {
```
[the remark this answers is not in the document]
brjeter: I don't liked "destination" because that's the name of a SAML element. That's why I went with "location". Maybe something like `getReceivingUrl`, `getRecipientUrl`, `getReceivingLocation`, etc?

library/src/main/java/org/codice/security/saml/EntityInformation.java
```diff
@@ -94,45 +90,38 @@ Binding getBinding(AuthnRequest request, Binding preferred) {
...
+
     if (request != null && request.getAssertionConsumerServiceURL() != null) {
-      return getAssertionConsumerServiceInfoByUrl(request);
+      si = getAssertionConsumerServiceInfoByUrl(request);
```
bakejeyner: ❓ Is there any reason for the ordering of `ByUrl`, `ByIndex`, and `ByProtocolBinding` when trying to obtain the `acsUrl`?

ctk/common/src/main/kotlin/org/codice/compliance/utils/TestCommon.kt
```diff
@@ -92,6 +93,7 @@ class TestCommon {
 
         val idpMetadata = Common.parseIdpMetadata()
         val SP_ISSUER = spMetadata.keys.first()
+        val SP_ENTITY_INFO = checkNotNull(spMetadata[SP_ISSUER])
 
         val acsUrl: Map<SamlProtocol.Binding, String?> by lazy {
             val spInfo = spMetadata[SP_ISSUER]
```
blen-desta: ✏️ We can use `SP_ENTITY_INFO` here instead of doing ` val spInfo = spMetadata[SP_ISSUER]`

deployment/distribution/src/main/resources/samlconf-sp-metadata.xml
```diff
@@ -22,7 +22,9 @@
     <md:SingleLogoutService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" Location="https://samlhost:8993/services/saml/logout/post"/>
     <md:SingleLogoutService Binding="urn:oasis:names:tc:SAML:2.0:bindings:SOAP" Location="https://samlhost:8993/services/saml/logout/soap"/>
     <md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect" Location="https://samlhost:8993/services/saml/sso/redirect" index="0"/>
-    <md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" Location="https://samlhost:8993/services/saml/sso/post" index="0"/>
-    <md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:2.0:bindings:PAOS" Location="https://samlhost:8993/services/saml/sso/paos" index="1"/>
+    <md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" Location="https://samlhost:8993/services/saml/sso/post" index="1"/>
```
blen-desta: ❗️ We also have to modify the config docker uses under `saml-conformance/deployment/docker/etc/org.codice.ddf.security.idp.server.IdpEndpoint.config`

library/src/main/java/org/codice/security/saml/EntityInformation.java
```diff
@@ -18,25 +18,18 @@
...
 import org.codice.security.saml.SamlProtocol.Binding;
 import org.opensaml.core.xml.schema.XSBase64Binary;
 import org.opensaml.saml.saml2.core.AuthnRequest;
-import org.opensaml.saml.saml2.metadata.AssertionConsumerService;
-import org.opensaml.saml.saml2.metadata.Endpoint;
-import org.opensaml.saml.saml2.metadata.EntityDescriptor;
-import org.opensaml.saml.saml2.metadata.KeyDescriptor;
-import org.opensaml.saml.saml2.metadata.SPSSODescriptor;
-import org.opensaml.saml.saml2.metadata.SingleLogoutService;
+import org.opensaml.saml.saml2.metadata.*;
```
bakejeyner: ❓ Are we allowing wildcard imports in java classes?

ctk/common/src/main/kotlin/org/codice/compliance/utils/RestAssuredExtensions.kt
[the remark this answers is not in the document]
bakejeyner: `getIntendedLocation`?

ctk/common/src/main/kotlin/org/codice/compliance/utils/RestAssuredExtensions.kt
brjeter: I see this extension function being used for any protocol which could use any binding, so I don't think we need to have an explicit check for artifact.

deployment/distribution/src/main/resources/samlconf-sp-metadata.xml
brjeter: So the additional ACS is no longer needed. I need to remove that. I am updating the indexes, however. It looks like the config files already as the updated indexes though 😱

library/src/main/java/org/codice/security/saml/EntityInformation.java
brjeter: Unimportant.

library/src/main/java/org/codice/security/saml/EntityInformation.java
```diff
@@ -147,32 +136,29 @@ public ServiceInfo getAssertionConsumerService(
   }
 
   private ServiceInfo getAssertionConsumerServiceInfoByUrl(AuthnRequest request) {
```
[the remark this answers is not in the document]
brjeter: I think I'd rather leave this as it is in DDF.

library/src/main/java/org/codice/security/saml/EntityInformation.java
brjeter: No. Obviously not. I can't believe you even had to ask :trollface:

blen-desta:
Hero'd successfully 👍

PR #128, opened 2018-05-02
MAINT - Removed duplicate methods for negative path testing

blen-desta:
Removes the duplicate methods for negative path testing and adds a method in `SAMLConformancException` to add the extra message on every thrown exception if it's an error test.

Error test exceptions will look like:

screen shot 2018-05-02 at 12 51 23 pm

#### Checklist:


library/src/main/kotlin/org/codice/compliance/SAMLComplianceException.kt
```diff
@@ -62,6 +66,18 @@ class SAMLComplianceException : Exception {
             }
         }
 
+        fun recreateExceptionWithErrorMessage(exception: SAMLComplianceException):
+            SAMLComplianceException {
+            return if (exception.cause != null) {
+                SAMLComplianceException(
+                    "${exception.message}$IDP_ERROR_RESPONSE_REMINDER_MESSAGE",
```
coyotesqrl: ❓ Would we like a space or a carriage return or some other break between these two strings?

library/src/main/kotlin/org/codice/compliance/SAMLComplianceException.kt
```diff
@@ -62,6 +66,18 @@ class SAMLComplianceException : Exception {
...
+                    exception.cause)
+            } else {
+                SAMLComplianceException(
+                    "${exception.message}$IDP_ERROR_RESPONSE_REMINDER_MESSAGE")
```
coyotesqrl: ❓ Would we like a space or a carriage return or some other break between these two strings?

library/src/main/kotlin/org/codice/compliance/SAMLComplianceException.kt
blen-desta: I had it as `"${exception.message}\n$IDP_ERROR_RESPONSE_REMINDER_MESSAGE"` but that was printing too many lines. ☝️(picture) is what it looks like now

brjeter:
Hero'd successfully 👍

PR #129, opened 2018-05-02
CORE 3.2.2 - Verify the response and assertions signature

blen-desta:
Validates the XML signatures
- The request's signature for POST
- The assertions' signatures for all bindings

#### Checklist:


ctk/common/src/main/kotlin/org/codice/compliance/verification/binding/RedirectBindingVerifier.kt
```diff
@@ -205,7 +207,7 @@ class RedirectBindingVerifier(httpResponse: Response) : BindingVerifier(httpResp
                 "encoding currently.")
 
         decodedMessage.debugPrettyPrintXml("Decoded SAML Response")
-        return Common.buildDom(decodedMessage)
+        return decodedMessage
```
brjeter: ✏️ Is there a reason we're not verifying the signature inside this method like in PostBindingVerifier?

library/src/main/java/org/codice/security/sign/SimpleSign.java
```diff
@@ -241,6 +252,76 @@ public boolean validateSignature(
...
+      throw new SignatureException("Can't get X509Certificate or PublicKey to verify signature.");
+    }
+
+    ClassLoader threadLoader = null;
+    try {
+      threadLoader = Thread.currentThread().getContextClassLoader();
+      Thread.currentThread()
+          .setContextClassLoader(
+              ApacheSantuarioSignatureValidationProviderImpl.class.getClassLoader());
+      SignatureValidator.validate(signature, credential);
```
brjeter: It looks like we're relying a lot on OpenSAML to do this signature validation. Could we similarly rely on OpenSAML in the `validateSignature` that already existed? If that's a possibility, I think we should maybe make a maintenance task for that.

ctk/common/src/main/kotlin/org/codice/compliance/verification/binding/RedirectBindingVerifier.kt
blen-desta: I forgot to change the post one. I didn't want to hide this in the decode. That way the decode is doing more than what the name says.

library/src/main/resources/SAMLSpecRefMessage.properties
```diff
@@ -378,3 +378,16 @@ SAMLBindings_3_5_4_d=The action attribute of the form MUST be the recipient's HT
...
   agent to deliver the message.
+
+#------------------
+# GENERAL
+#------------------
+SAMLGeneral.doc=SAML V2.0 Standard
+
+SAMLGeneral.uri=https://wiki.oasis-open.org/security/FrontPage
+
+SAMLGeneral_a=An error occurred while validating the signature. See section 3.4.4.1 of the Binding document \
```
bakejeyner: ❗️ Line length

library/src/main/java/org/codice/security/sign/SimpleSign.java
blen-desta: We have one to investigate OpenSAMLlibraries in general to simplify a lot of the things we do

blen-desta:
aborting test
build now

library/src/main/java/org/codice/security/sign/SimpleSign.java
```diff
@@ -241,6 +252,76 @@ public boolean validateSignature(
...
+    BasicX509Credential credential = null;
+    if (samlKeyInfo.getCerts() != null) {
+      credential = new BasicX509Credential(samlKeyInfo.getCerts()[0]);
+    } else {
+      throw new SignatureException("Can't get X509Certificate or PublicKey to verify signature.");
+    }
+
+    ClassLoader threadLoader = null;
+    try {
+      threadLoader = Thread.currentThread().getContextClassLoader();
```
bakejeyner: ❓ Not really a review question, but why does DDF change the threads `contextClassLoader` here? Just want to know in case it's important or completely unnecessary.

library/src/main/java/org/codice/security/sign/SimpleSign.java
```diff
@@ -241,6 +252,76 @@ public boolean validateSignature(
     }
   }
 
+  public void validateSignature(Signature signature) throws SignatureException {
```
bakejeyner: ❗️ Should add a comment about further researching this method. From what I've seen it looks incomplete.

For instance: If no `KeyInfo` is passed in, this method errors out. The spec says that if no `KeyInfo` exists then we should try and decrypt it with their signing key.

library/src/main/resources/SAMLSpecRefMessage.properties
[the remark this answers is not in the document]
blen-desta: Doesn't detekt catch this?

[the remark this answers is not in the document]
brjeter: It does in my experience

ctk/common/src/main/kotlin/org/codice/compliance/verification/binding/BindingVerifier.kt
```diff
@@ -37,6 +43,32 @@ abstract class BindingVerifier(val httpResponse: Response) {
...
+            } catch (e: SimpleSign.SignatureException) {
+                throw SAMLComplianceException.create(SAMLGeneral_b,
+                    message = "Invalid signature.\n${e.message}",
+                    cause = e)
+            }
+        }
+
+        /** Creates an OpenSAML Response Object */
+        private fun buildResponseObject(samlResponse: String):
+            org.opensaml.saml.saml2.core.Response {
```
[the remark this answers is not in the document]
blen-desta: I'm gonna keep it as is cause there are just two usages of it. In my opinion, that's a bit confusing but we can change it later if there are more usages.

library/src/main/java/org/codice/security/sign/SimpleSign.java
blen-desta: Not Sure SimpleSign is taken from DDF.

library/src/main/java/org/codice/security/sign/SimpleSign.java
```diff
@@ -241,6 +252,76 @@ public boolean validateSignature(
...
+    }
+  }
+
+  private void validateSignatureAndSamlKey(Signature signature, SAMLKeyInfo samlKeyInfo)
+      throws SignatureException {
+    SAMLSignatureProfileValidator validator = new SAMLSignatureProfileValidator();
+    try {
+      validator.validate(signature);
+    } catch (org.opensaml.xmlsec.signature.support.SignatureException e) {
+      throw new SignatureException("Error validating the SAMLKey signature", e);
```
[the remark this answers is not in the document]
blen-desta: Not Sure SimpleSign is taken from DDF. I can change that error message

library/src/main/java/org/codice/security/sign/SimpleSign.java
blen-desta: There is a card to investigate that.

ctk/common/src/main/kotlin/org/codice/compliance/verification/binding/BindingVerifier.kt
```diff
@@ -45,9 +43,10 @@ abstract class BindingVerifier(val httpResponse: Response) {
         }
 
         /** Verifies the response's and assertions' signatures */
-        fun verifyXmlSignature(samlResponse: String) {
+        fun verifyXmlSignature(dom: Document) {
```
bakejeyner: ✏️ Since this method verifies more than 1 signature, it might make more sense to name it plural: `verifyXmlSignatures`.

blen-desta:
build now

PR #130, opened 2018-05-02
MAINT - Re-enable error tests

brjeter:
I couldn't run these even with the `-e` option when heroing Blen's PR. This re-enables them (though they still won't run from the script without `-e`).
#### Checklist:


blen-desta:
❗️ We need to remove the `-e` from `saml-conformance/deployment/docker/wait.sh` if we are re-enabling all of them. Our CI build will run them.

brjeter:
Yeah I'll address those warnings as part of this ass well.

blen-desta:
build now

PR #134, opened 2018-05-03
MAINT - Adds unit tests for CoreVerifierSpec and does some small cleanup

coyotesqrl:
#### Checklist:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,7 +25,7 @@ description = 'SAML Conformance Test Kit'
 
 project.ext {
     project_version = '1.0-SNAPSHOT'
-    kotlin_test_version = '3.1.0-RC2'
+    kotlin_test_version = '3.1.4'
     rest_assured_version = '2.9.0'
     org_slf4j_version = '1.7.1'
     guava_version = '24.0-jre'
--- a/build.gradle
+++ b/build.gradle
@@ -25,7 +25,7 @@ description = 'SAML Conformance Test Kit'
 
 project.ext {
     project_version = '1.0-SNAPSHOT'
-    kotlin_test_version = '3.1.0-RC3'
+    kotlin_test_version = '3.1.0-RC2'
     rest_assured_version = '2.9.0'
     org_slf4j_version = '1.7.1'
     guava_version = '24.0-jre'
--- a/ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
+++ b/ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
@@ -45,12 +45,6 @@ abstract class CoreVerifier(protected val node: Node) {
                                   expectedStatusCode: String) {
             SchemaValidator.validateSAMLMessage(node)
             val status = node.children("Status")
-            if (status.size != 1)
-                throw SAMLComplianceException.createWithPropertyMessage(samlErrorCode,
-                        property = "Status",
-                        actual = "[]",
-                        expected = "<Status>",
-                        node = node)
             status[0].children(STATUS_CODE).forEach {
                 val code = it.attributeText("Value")
 
--- a/ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
+++ b/ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
@@ -88,11 +88,17 @@ abstract class CoreVerifier(protected val node: Node) {
 
             val notBeforeValue = Instant.parse(notBefore.textContent)
             val notOnOrAfterValue = Instant.parse(notOnOrAfter.textContent)
-            if (notBeforeValue.isAfter(notOnOrAfterValue))
+            if (notBeforeValue == notOnOrAfterValue || notBeforeValue.isAfter(notOnOrAfterValue))
                 throw SAMLComplianceException.create(samlCode,
                         message = "NotBefore element with value $notBeforeValue is not less " +
                                 "than NotOnOrAfter element with value $notOnOrAfterValue.",
                         node = node)
+        }
+
+        private fun retrieveCurrentEncryptedElements(responseDom: Node): List<Node> {
+            return responseDom.recursiveChildren("EncryptedAssertion") +
+                    responseDom.recursiveChildren("EncryptedAttribute") +
+                    responseDom.recursiveChildren("EncryptedID")
         }
     }
 
@@ -100,7 +106,7 @@ abstract class CoreVerifier(protected val node: Node) {
      * Verify response against the Core Spec document
      */
     open fun verify() {
-        preProcess(node)
+        preProcess()
         verifyCommonDataType(node)
         SamlAssertionsVerifier(node).verify()
         SamlVersioningVerifier(node).verify()
@@ -111,12 +117,11 @@ abstract class CoreVerifier(protected val node: Node) {
     open fun verifyEncryptedElements() {
     }
 
-    private fun preProcess(responseDom: Node,
-                           encVerifier: EncryptionVerifier = EncryptionVerifier()) {
-        val encElements = retrieveCurrentEncryptedElements(responseDom)
+    private fun preProcess(encVerifier: EncryptionVerifier = EncryptionVerifier()) {
+        val encElements = retrieveCurrentEncryptedElements(node)
         if (encElements.isEmpty()) {
             Log.debugWithSupplier {
-                "Decrypted SAML Response:\n\n ${responseDom.prettyPrintXml()}"
+                "Decrypted SAML Response:\n\n ${node.prettyPrintXml()}"
             }
             return
         }
@@ -126,15 +131,9 @@ abstract class CoreVerifier(protected val node: Node) {
                     " on the SAML Response."
         }
 
-        SchemaValidator.validateSAMLMessage(responseDom)
+        SchemaValidator.validateSAMLMessage(node)
         verifyEncryptedElements()
         encVerifier.verifyAndDecryptElements(encElements)
-        preProcess(responseDom, encVerifier)
-    }
-
-    private fun retrieveCurrentEncryptedElements(responseDom: Node): List<Node> {
-        return responseDom.recursiveChildren("EncryptedAssertion") +
-                responseDom.recursiveChildren("EncryptedAttribute") +
-                responseDom.recursiveChildren("EncryptedID")
+        preProcess(encVerifier)
     }
 }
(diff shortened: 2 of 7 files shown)


ctk/common/src/test/kotlin/org/codice/compilance/verification/core/CoreVerifierSpec.kt
```diff
@@ -0,0 +1,122 @@
...
+            }
+        }
+
+        "failure with expected code of REQUESTER includes SAMLCore 3.2.1 response" {
+            buildDom(response("urn:samlconf:fail")).let {
+                shouldThrow<SAMLComplianceException> {
+                    CoreVerifier.verifyErrorStatusCode(it, specRefMsg, TestCommon.REQUESTER)
+                }.message?.shouldContain(SAMLCore_3_2_1_d.message)
+            }
+        }
```
brjeter: ❗️ We should have a test for multiple status elements in the response.

ctk/common/src/test/kotlin/org/codice/compilance/verification/core/CoreVerifierSpec.kt
```diff
@@ -0,0 +1,122 @@
...
+
+        "mismatched status code fails and does not contain SAMLCore 3.2.1 response" {
+            buildDom(response("urn:samlconf:fail")).let {
+                shouldThrow<SAMLComplianceException> {
+                    CoreVerifier.verifyErrorStatusCode(it, specRefMsg, "urn:samlconf:expected")
+                }.message?.shouldNotContain(SAMLCore_3_2_1_d.message)
+            }
+        }
+
+        "failure with expected code of REQUESTER includes SAMLCore 3.2.1 response" {
```
brjeter: ✏️ Super nit: To be more consistent with the similar test above. "mismatch with expected code of REQUESTER contains SAMLCore 3.2.1. response".

ctk/common/src/test/kotlin/org/codice/compilance/verification/core/CoreVerifierSpec.kt
```diff
@@ -0,0 +1,122 @@
...
+                }
+            }
+
+            buildDom("""<fld NotBefore="$later" NotOnOrAfter="$later"/>""").let {
+                shouldThrow<SAMLComplianceException> {
+                    CoreVerifier.validateTimeWindow(it, specRefMsg)
+                }
+            }
+        }
+    }
```
brjeter: ❗️ Technically you might want to include the negative path of having only one of `NotBefore` or `NotOnOrAfter` and the value of it being invalid. You could assert that the SamlComplianceException shouldNotContain the specRefMsg since `CommonDataTypeVerifier.verifyDateTimeValues(this)` will not include the passed in message.

ctk/common/src/test/kotlin/org/codice/compilance/verification/core/CoreVerifierSpec.kt
```diff
@@ -0,0 +1,122 @@
...
+}
```
brjeter: ❗️ I get not verifying that the CoreVerifier delegates to the other Verifier objects. That sounds like a pain and would require significant changes. But shouldn't we be at least testing the other two pieces of functionality in this class? `preProcess` and `verifyEntityIdentifiers`? I'd be fine with making those functions internal to make that possible.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
```diff
@@ -77,6 +77,36 @@ abstract class CoreVerifier(protected val node: Node) {
...
+            val notBeforeValue = Instant.parse(notBefore.textContent)
+            val notOnOrAfterValue = Instant.parse(notOnOrAfter.textContent)
+            if (notBeforeValue == notOnOrAfterValue || notBeforeValue.isAfter(notOnOrAfterValue))
+                throw SAMLComplianceException.create(samlCode,
+                        message = "NotBefore element with value $notBeforeValue is not less " +
+                                "than NotOnOrAfter element with value $notOnOrAfterValue.",
+                        node = node)
+        }
+
+        private fun retrieveCurrentEncryptedElements(responseDom: Node): List<Node> {
```
bakejeyner: ❓ Is there a reason we are moving this into the companion?

ctk/common/src/test/kotlin/org/codice/compilance/verification/core/CoreVerifierSpec.kt
bakejeyner: I just had a realization from reading this comment that I wanted to share (correct me if I'm wrong 😨):

It would be overkill / bad practice to add specific exception testing for the delegated methods. For example: making sure that when we provide a bad data type, we get `SAMLCore_1_3_3_a` in the exception message.

However, if we know that those methods should always throw an exception, catching a general exception would be ideal for testing that the method will error out. That way we can change the other class's exception as much as necessary without needing to changing this seemingly unrelated test class.

brjeter: Yeah it would not be a very good practice, that's true. Ideally we'd be able to verify that the method we're testing ends up calling `CommonDataTypeVerifier.verifyDateTimeValues(this)`, but it being a static method call makes that difficult. I guess expecting any SamlComplianceException would be fine.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
blen-desta: It has always been in the companion object

bakejeyner: I think it was originally an instance method.

blen-desta: Nope pretty sure it's always been in the companion object. https://github.com/connexta/saml-conformance/blob/master/ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt#L111

coyotesqrl:
I still have yet to address the comments on the PR, but at least it now runs correctly again.

ctk/common/src/test/kotlin/org/codice/compilance/verification/core/CoreVerifierSpec.kt
coyotesqrl: I'm not sure what you mean by having only one and it not being valid. Having only one gets tested in several permutations in `validateTimeWindow with null datetime attributes` and behaves as expected.

ctk/common/src/test/kotlin/org/codice/compilance/verification/core/CoreVerifierSpec.kt
coyotesqrl: `preProcess` basically tests:

1. Schema validation (outside the scope of this test class)
2. `verifyEncryptedElements`, which is implemented in the subclasses of `CoreVerifier` and should be tested against those subclasses
3. The functioning of the `EncryptionVerifier` (outside the scope of this test class)

Unless there's something else in there you think should be tested, I'm not sure how to dive on that.

`verifyEntityIdentifiers` is OBE, as it's been moved out of this class.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
```diff
@@ -45,12 +45,6 @@ abstract class CoreVerifier(protected val node: Node) {
                                   expectedStatusCode: String) {
             SchemaValidator.validateSAMLMessage(node)
             val status = node.children("Status")
-            if (status.size != 1)
```
bakejeyner: ❓ Was this taken out because you can have more than 1 status, or because Schema Validation will catch it?

ctk/common/src/test/kotlin/org/codice/compilance/verification/core/CoreVerifierSpec.kt
```diff
@@ -0,0 +1,164 @@
...
+                |</s:Response>
+                """.trimMargin()
+        }
+
+        val specRefMsg = mockk<SAMLCoreRefMessage>()
+        every { specRefMsg.message } returns "test message"
+        every { specRefMsg.name } returns "TEST"
+
+        "single error status response has expected code" {
+            val statusCode = "urn:samlconf:fail"
```
bakejeyner: ✏️ Could make statusCode constants, for example
```
private const val statusCodeFail = "urn:samlconf:fail"
private const val statusCodeExpected = "urn:samlconf:expected"
...
```

ctk/common/src/test/kotlin/org/codice/compilance/verification/core/CoreVerifierSpec.kt
```diff
@@ -0,0 +1,164 @@
...
+import java.time.Duration
+import java.time.Instant
+
+@Suppress("StringLiteralDuplication", "MagicNumber")
+class CoreVerifierSpec : StringSpec() {
+    init {
+        val now = Instant.now()
+        val later = now.plus(Duration.ofMinutes(30))
+
+        val response = { statusCode: String ->
```
bakejeyner: ✏️ If you use methods instead of closures, you can provide default values for the parameters. A use case could be that instead of passing `"urn:samlconf:fail"` to most of these closures, you can just call the method with not arguments.

build.gradle
```diff
@@ -25,7 +25,7 @@ description = 'SAML Conformance Test Kit'
 
 project.ext {
     project_version = '1.0-SNAPSHOT'
-    kotlin_test_version = '3.1.0-RC3'
+    kotlin_test_version = '3.1.0-RC2'
```
blen-desta: ✏️ We can change this to `3.1.4`

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CoreVerifier.kt
brjeter: Schema validation will catch it.

ctk/common/src/test/kotlin/org/codice/compilance/verification/core/CoreVerifierSpec.kt
brjeter: My main concern is that we could have a bug in the flow of logic in the `preProcess` method. Or we could introduce one in the future. That being said, it would be a step above a purely "unit" test. Maybe the fact that we can't test this without calling out to a lot of external classes means the method should be moved or redesigned 🤔

coyotesqrl:
build now

PR #135, opened 2018-05-07
Core-3.4.1.4 Add subject comparison to verifiers

bakejeyner:

* Adds calls to the already in place subject comparison framework. Refactors the framework a bit to work with openSaml objects.

* Adds an additional MUST verification from Core 3.4.1.4: `The assertion(s) returned MUST contain a <saml:Subject> element that represents the presenter. The identifier type and format are determined by the identity provider.`. **Note** that some of this MUST is added to `NotTested.md`.

* Adds extra status code checks to the SSO test classes. Before these checks, if an invalid response status was found, the `RestAssuredExtensions` `getBindingVerifier` method would throw a deceiving `UnsupportedOperationException` instead of an exception explaining the invalid http status code.

* Adds a section to the base `Readme` mentioning our coverage documentation.

#### Checklist:


ctk/common/src/main/kotlin/org/codice/compliance/verification/binding/PostBindingVerifier.kt
```diff
@@ -35,7 +35,6 @@ import kotlin.test.assertNotNull
 class PostBindingVerifier(httpResponse: Response) : BindingVerifier(httpResponse) {
     /** Verify the response for a post binding */
     override fun decodeAndVerify(): Node {
-        verifyHttpStatusCode(httpResponse.statusCode)
```
bakejeyner: Taking this out of the verify method because it will always be called directly from the tests.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
```diff
@@ -99,16 +103,16 @@ class SubjectComparisonVerifier(private val request: Node? = null, private val r
     @Suppress("NestedBlockDepth" /* Simple `let` nesting */)
     fun verifySubjectsMatchAuthnRequest() {
 
-        val requestSubject = request?.children(SUBJECT)?.firstOrNull() ?: return
+        val requestSubject = samlRequest?.dom?.children(SUBJECT)?.firstOrNull() ?: return
```
bakejeyner: I chose to just pull out the dom here instead of using the openSaml objects because we haven't yet figured out if we want to majorly use openSaml yet.

README.md
```diff
@@ -77,6 +77,12 @@ To build a docker image, execute `gradlew build docker`.
 >
 > Docker is used exclusively for our Jenkins builds.
 
+### SAML Specification Coverage
+Documentation on positive path IdP SSO coverage can be found in:
```
blen-desta: ✏️ This is a duplicate of what is said on line 99.

> The [coverage](ctk/idp/coverage) directory is used to track which sections of each SAML specification are covered by these tests.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/responses/CoreAuthnRequestProtocolVerifier.kt
```diff
@@ -68,15 +72,15 @@ class CoreAuthnRequestProtocolVerifier(private val authnRequest: AuthnRequest,
                     node = samlResponseDom)
 
         if (assertions.all { it.children(AUTHN_STATEMENT).isEmpty() })
-            throw SAMLComplianceException.create(SAMLCore_3_4_a, SAMLCore_3_4_1_4_c,
+            throw SAMLComplianceException.create(SAMLCore_3_4_a, SAMLCore_3_4_1_4_d,
```
bakejeyner: These renames are because I added a new `SAMLCore_3_4_1_4_c`.

ctk/common/src/main/kotlin/org/codice/compliance/verification/binding/PostBindingVerifier.kt
blen-desta: ✏️ I think we should delete it from the tests and leave it here. For me, it makes sense for the Binding verifier to call it versus each test. If the issue is the `UnsupportedOperationException` couldn't we catch that and throw a `SAMLComplianceException`?

ctk/common/src/main/kotlin/org/codice/compliance/verification/profile/SingleSignOnProfileVerifier.kt
```diff
@@ -57,6 +58,7 @@ class SingleSignOnProfileVerifier(private val authnRequest: AuthnRequest,
     fun verify() {
         verifyIssuer()
         verifySsoAssertions()
+        SubjectComparisonVerifier(authnRequest, samlResponseDom).verifySubjectsMatchSSO()
```
blen-desta: ❓ Shouldn't you pass in `null` cause we're comparing the Response's Subjects? If not we won't need the `request: Node? = null` in the `SubjectComparisonVerifier`'s constructor

ctk/idp/coverage/Core.md
```diff
@@ -85,7 +85,7 @@ Unmarked sections need attention
 -			3.4.1.2 Element <Scoping>
 -			3.4.1.3 Element <IDPList>
 -				3.4.1.3.1 Element <IDPEntry>
-			3.4.1.4 Processing Rules
++			3.4.1.4 Processing Rules
```
blen-desta: 💯

ctk/idp/coverage/Profiles.md
```diff
@@ -33,7 +33,7 @@ Unmarked sections need attention
 -			4.1.3.6 Service Provider Grants or Denies Access to User Agent
 		4.1.4 Use of Authentication Request Protocol
 			4.1.4.1 <AuthnRequest> Usage
-			4.1.4.2 <Response> Usage
++			4.1.4.2 <Response> Usage
```
blen-desta: 💯

ctk/common/src/main/kotlin/org/codice/compliance/verification/binding/PostBindingVerifier.kt
bakejeyner: A problem with that approach is that if they actually do have a binding we don't support, we may catch that error and throw a misleading "Incorrect http status code" `SAMLComplianceException`.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
brjeter: Then why did we change the method to take an opensaml object in the first place? 😂

bakejeyner: To pass around the dom :trollface:

bakejeyner:
build now

PR #139, opened 2018-05-09
Core 8.3.7 & 8.3.8 - Persistent and Transient Identifier

blen-desta:

- 8.3.7 Persistent Identifier
- 8.3.8 Transient Identifier

#### Checklist:

### DIFF
--- a/ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlDefinedIdentifiersVerifier.kt
+++ b/ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlDefinedIdentifiersVerifier.kt
@@ -19,11 +19,22 @@ import org.codice.compliance.SAMLCore_8_2_2_a
 import org.codice.compliance.SAMLCore_8_2_3_a
 import org.codice.compliance.SAMLCore_8_3_6_a
 import org.codice.compliance.SAMLCore_8_3_6_b
+import org.codice.compliance.SAMLCore_8_3_7_a
+import org.codice.compliance.SAMLCore_8_3_7_b
+import org.codice.compliance.SAMLCore_8_3_7_c
+import org.codice.compliance.SAMLCore_8_3_7_d
+import org.codice.compliance.SAMLCore_8_3_8_a
 import org.codice.compliance.attributeNode
 import org.codice.compliance.attributeText
 import org.codice.compliance.children
 import org.codice.compliance.recursiveChildren
 import org.codice.compliance.utils.TestCommon
+import org.codice.compliance.utils.TestCommon.Companion.FORMAT
+import org.codice.compliance.utils.TestCommon.Companion.PERSISTENT_ID
+import org.codice.compliance.utils.TestCommon.Companion.SP_ISSUER
+import org.codice.compliance.utils.TestCommon.Companion.SP_NAME_QUALIFIER
+import org.codice.compliance.utils.TestCommon.Companion.TRANSIENT_ID
+import org.codice.compliance.utils.TestCommon.Companion.idpMetadata
 import org.w3c.dom.DOMException
 import org.w3c.dom.Node
 import java.net.URI
@@ -34,6 +45,7 @@ internal class SamlDefinedIdentifiersVerifier(val node: Node) {
 
     companion object {
         private const val ENTITY_ID_MAX_LEN = 1024
+        private const val ID_VALUE_LENGTH_LIMIT = 256
         private const val ATTRIBUTE_NAME_FORMAT_URI =
                 "urn:oasis:names:tc:SAML:2.0:attrname-format:uri"
         private const val ATTRIBUTE_NAME_FORMAT_BASIC =
@@ -49,6 +61,8 @@ internal class SamlDefinedIdentifiersVerifier(val node: Node) {
     fun verify() {
         verifyActionNamespaceIdentifiers()
         verifyAttributeNameFormatIdentifiers()
+        verifyPersistentIdentifiers()
+        verifyTransientIdentifiers()
         verifyEntityIdentifiers()
     }
 
@@ -132,7 +146,7 @@ internal class SamlDefinedIdentifiersVerifier(val node: Node) {
 
     private fun checkEntityIdentifier(node: Node) {
         if (node.attributeNode("NameQualifier") != null ||
-            node.attributeNode("SPNameQualifier") != null ||
+            node.attributeNode(SP_NAME_QUALIFIER) != null ||
             node.attributeNode("SPProvidedID") != null) {
             throw SAMLComplianceException.create(SAMLCore_8_3_6_a,
                 message = "No Subject element found.",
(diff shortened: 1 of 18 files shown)


ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlDefinedIdentifiersVerifier.kt
```diff
@@ -118,4 +130,48 @@ internal class SamlDefinedIdentifiersVerifier(val node: Node) {
...
+                        node = it)
+
+                if (it.children("NameQualifier")
+                        .any { it.textContent != idpMetadata.entityId })
+                    throw SAMLComplianceException.create(SAMLCore_8_3_7_b,
+                        SAMLCore_8_3_7_c,
+                        message = "The Persistent ID's NameQualifier isn't ${idpMetadata.entityId}",
+                        node = it)
+
+                if (it.children("SPNameQualifier")
```
brjeter: ❗️ Aren't `NameQualifier` and `SPNameQualifier` attributes rather than child elements?
✏️ If the above is true: When fixing that, you could put the actual attributeText in the exception message instead of just putting the expected value.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlDefinedIdentifiersVerifier.kt
```diff
@@ -118,4 +130,48 @@ internal class SamlDefinedIdentifiersVerifier(val node: Node) {
...
+            .filter { it.namespaceURI == TRANSIENT_ID }
+            .filter { it.textContent != null }
+            .forEach {
+                if (it.textContent.length > ID_VALUE_LENGTH_LIMIT)
+                    throw SAMLComplianceException.create(SAMLCore_8_3_8_a,
+                        message = "The length of the Transient ID's value was greater than " +
+                            "$ID_VALUE_LENGTH_LIMIT characters.",
+                        node = it)
+
+                CommonDataTypeVerifier.verifyIdValues(it, SAMLCore_8_3_8_a)
```
brjeter: ✏️ This is out of the scope of this PR, but why is this method called `verifyIdValues` with an "s"? Shouldn't it be singular?

brjeter: ❓ Is there a reason we're not making the `CommonDataType` call for persistent identifiers?

blen-desta: That's because the Transient explicitly says: `Transient identifier values MUST be generated in accordance with the rules for SAML identifiers (see Section 1.3.4)`

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlDefinedIdentifiersVerifier.kt
blen-desta: Correct. Fixed

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlDefinedIdentifiersVerifier.kt
```diff
@@ -118,4 +130,48 @@ internal class SamlDefinedIdentifiersVerifier(val node: Node) {
...
+            .filter { it.namespaceURI == PERSISTENT_ID }
+            .forEach {
+                if (it.textContent != null && it.textContent.length > ID_VALUE_LENGTH_LIMIT)
+                    throw SAMLComplianceException.create(SAMLCore_8_3_7_a,
+                        message = "The length of the Persistent ID's value was greater than " +
+                            "$ID_VALUE_LENGTH_LIMIT characters.",
+                        node = it)
+
+                if (it.children("NameQualifier")
+                        .any { it.textContent != idpMetadata.entityId })
```
bakejeyner: ❗️ What if the IdP we are testing got an assertion from another SAML party. Then the NameQualifier wouldn't be the EntityID from the metadata. See `SAMLCore_8_3_7_b` and `SAMLCore_8_3_7_c`.

library/src/main/resources/SAMLSpecRefMessage.properties
```diff
@@ -302,6 +302,22 @@ SAMLCore_8_3_6_b=[T]he content of the element is the identifier of an entity tha
   SAML profiles (such as a service provider supporting the browser SSO profile)...The syntax of \
   such an identifier is a URI of not more than 1024 characters in length.
 
+SAMLCore_8_3_7_a=Persistent name identifier values MUST NOT exceed a length of 256 characters.
+
+SAMLCore_8_3_7_b=The element's NameQualifier attribute, if present, MUST contain the unique \
```
bakejeyner: ❗️ We should add this to `NotTested`.

In the reasoning we should note that we can't tell if a message was generated from another SAML entity.

library/src/main/resources/SAMLSpecRefMessage.properties
```diff
@@ -302,6 +302,22 @@ SAMLCore_8_3_6_b=[T]he content of the element is the identifier of an entity tha
...
+  identifier of the identity provider that generated the identifier (see Section 8.3.6).
+
+SAMLCore_8_3_7_c=Note that a different system entity might later issue its own protocol message or \
```
bakejeyner: ❗️ We should add this to `NotTested`.

In the reasoning we should note that we can't tell if a message was generated from another SAML entity, or if the attribute was omitted all together.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlDefinedIdentifiersVerifier.kt
```diff
@@ -118,4 +130,46 @@ internal class SamlDefinedIdentifiersVerifier(val node: Node) {
...
+
+                if (it.attributeText("NameQualifier") != idpMetadata.entityId )
+                    throw SAMLComplianceException.create(SAMLCore_8_3_7_b,
+                        SAMLCore_8_3_7_c,
+                        message = "The Persistent ID's NameQualifier isn't ${idpMetadata.entityId}",
+                        node = it)
+
+                if (it.attributeText("SPNameQualifier") != SP_ISSUER )
+                    throw SAMLComplianceException.create(SAMLCore_8_3_7_d,
+                        message = "The Persistent ID's SPNameQualifier isn't $SP_ISSUER",
```
brjeter: ✏️ You could put the actual value in the exception message. I've used `.let` in the past to reuse the attribute's text value before.
```
it.attributText("SPNameQualifier").let {
        if (it != SP_ISSUER )
                    throw SAMLComplianceException.create(SAMLCore_8_3_7_d,
                        message = "The Persistent ID's SPNameQualifier [$it] isn't equal to $SP_ISSUER",
                        node = it)
}
```

library/src/main/resources/SAMLSpecRefMessage.properties
blen-desta: Not sure what you mean? It says that the `NameQualifier` needs to have the IdP's entity ID.

library/src/main/resources/SAMLSpecRefMessage.properties
blen-desta: That was being added in a separate PR. I guess I can do it in this one

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlDefinedIdentifiersVerifier.kt
blen-desta: It says the entity id of the `identity provider`

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlDefinedIdentifiersVerifier.kt
```diff
@@ -146,4 +160,50 @@ internal class SamlDefinedIdentifiersVerifier(val node: Node) {
...
+
+    /** 8.3.8 Transient Identifier */
+    private fun verifyTransientIdentifiers() {
+        node.recursiveChildren()
+            .filter { it.attributeText(FORMAT) == TRANSIENT_ID }
+            .filter { it.textContent != null }
+            .forEach {
+                if (it.textContent.length > ID_VALUE_LENGTH_LIMIT)
+                    throw SAMLComplianceException.create(SAMLCore_8_3_8_a,
+                        message = "The length of the Transient ID's value was greater than " +
```
bakejeyner: ✏️ Could add the length to the message:
```
"The length of the Transient ID's value [${it.textContent.length} characters] 
was greater than $ID_VALUE_LENGTH_LIMIT characters."
```

blen-desta:
build now

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlDefinedIdentifiersVerifier.kt
```diff
@@ -146,4 +160,52 @@ internal class SamlDefinedIdentifiersVerifier(val node: Node) {
...
+                            "characters.",
+                        node = it)
+
+                it.attributeText(SP_NAME_QUALIFIER)?.let { nameQualifier ->
+                    if (nameQualifier != idpMetadata.entityId)
+                        throw SAMLComplianceException.create(SAMLCore_8_3_7_b,
+                            SAMLCore_8_3_7_c,
+                            message = "The Persistent ID's NameQualifier [$nameQualifier] is not " +
+                                "equal to ${idpMetadata.entityId}",
+                            node = it)
```
brjeter: When I was looking over my comments, I realized my code was incorrect because it was still passing in "it" for the node which wasn't the node anymore. But you got it 💯

PR #140, opened 2018-05-09
Core-8.3.2 Email address name identifier format

bakejeyner:
#### Checklist:


ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlDefinedIdentifiersVerifier.kt
```diff
@@ -124,6 +134,21 @@ internal class SamlDefinedIdentifiersVerifier(val node: Node) {
         }
     }
 
+    /** 8.3 Name Identifier Format Identifiers */
+    private fun verifyNameIdentifierFormatIdentifiers() {
+        node.recursiveChildren()
+                .filter { it.attributeText("Format") == EMAIL_URI }
+                .forEach {
```
bakejeyner: I didn't use `.any` here because I wanted to use the actual Identifier in the error message.

blen-desta:
Now that we have time. I think this is worth unit testing

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlDefinedIdentifiersVerifier.kt
```diff
@@ -44,69 +43,50 @@ import javax.xml.parsers.DocumentBuilderFactory
...
+
+        internal val RWEDC_URI_LIST = listOf(
                 "urn:oasis:names:tc:SAML:1.0:action:rwedc-negation",
                 "urn:oasis:names:tc:SAML:1.0:action:rwedc"
         )
     }
 
     /** 8 SAML-Defined Identifiers */
     fun verify() {
-        verifyActionNamespaceIdentifiers()
```
blen-desta: ❓ What happened to `verifyActionNamespaceIdentifiers`?

ctk/common/src/test/kotlin/org/codice/compilance/verification/core/SamlDefinedIdentifiersVerifierSpec.kt
```diff
@@ -0,0 +1,281 @@
...
+        /* 8.3 Name Identifier Format Identifiers */
+        "valid Email name identifier" {
+            createResponse(
+                    identifierValue = "example-email@domain.com",
+                    identifierFormat = NAME_ID_FORMAT_EMAIL).let {
+                SamlDefinedIdentifiersVerifier(it).verify()
+            }
+        }
+
+        "invalid Email name identifier (multiple '@'s)" {
```
blen-desta: You should combine all 9 invalid email tests into one
```
"invalid Email name identifier" {
            forAll("example@email@domain.com", "example-email.domain.com" etc ....) { email ->
                createResponse(
                     identifierValue = email,
                     identifierFormat = NAME_ID_FORMAT_EMAIL).let {
                 shouldThrow<SAMLComplianceException> {
                     SamlDefinedIdentifiersVerifier(it).verify()
                 }.message?.shouldContain(SAMLCore_8_3_2_a.message)
             }
        }
```

You can also add all the email to a list and do `forAll(invalidEmailList) { email ->`

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlDefinedIdentifiersVerifier.kt
```diff
@@ -44,69 +43,50 @@ import javax.xml.parsers.DocumentBuilderFactory
...
-
     /** 8.2 URI/Basic name attribute formats */
     private fun verifyAttributeNameFormatIdentifiers() {
         node.recursiveChildren("Attribute").forEach {
-            val name = it.attributeNode("Name")
-            val nameFormatText = it.attributeText("NameFormat")
-            if (name == null || nameFormatText == null) {
-                return
-            }
+            val name = it.attributeNode("Name") ?: return
```
brjeter: ✏️ You could just do `it.attributeText("Name")`.

ctk/common/src/test/kotlin/org/codice/compilance/verification/core/SamlDefinedIdentifiersVerifierSpec.kt
```diff
@@ -0,0 +1,281 @@
...
+                    |    Format="$identifierFormat">$identifierValue</saml:Issuer>
+                    |    <saml:AttributeStatement>
+                    |      <saml:Attribute Name="$attributeName" NameFormat="$attributeFormat" />
+                    |    </saml:AttributeStatement>
+                    |  </saml:Assertion>
+                    |</samlp:Response>
+                    """.trimMargin())
+        }
+
+        /* 8.2 URI/Basic name attribute formats */
```
brjeter: ✏️ Do we also want to test when either the name or nameFormat attributes are `null`?

ctk/common/src/test/kotlin/org/codice/compilance/verification/core/SamlDefinedIdentifiersVerifierSpec.kt
```diff
@@ -0,0 +1,281 @@
...
+            createResponse(
+                    attributeName = "Whitespace Not Allowed In Attribute Name",
+                    attributeFormat = ATTRIBUTE_NAME_FORMAT_BASIC).let {
+                shouldThrow<SAMLComplianceException> {
+                    SamlDefinedIdentifiersVerifier(it).verify()
+                }.message?.shouldContain(SAMLCore_8_2_3_a.message)
+            }
+        }
+
+        /* 8.3 Name Identifier Format Identifiers */
```
brjeter: ✏️ Should read like `/** 8.3.2 Email Address */` to be consistent with the actual class.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlDefinedIdentifiersVerifier.kt
```diff
@@ -44,69 +43,50 @@ import javax.xml.parsers.DocumentBuilderFactory
...
     }
 
     /** 8 SAML-Defined Identifiers */
     fun verify() {
-        verifyActionNamespaceIdentifiers()
         verifyAttributeNameFormatIdentifiers()
+        verifyEmailAddressIdentifier()
         verifyPersistentIdentifiers()
         verifyTransientIdentifiers()
         verifyEntityIdentifiers()
```
brjeter: This is not a high enough priority to actually do anything about, but I'm just realizing that we recurse the entire dom for each of these verify method calls when we could just recurse once and have a big `when` statement for the different formats.

ctk/common/src/test/kotlin/org/codice/compilance/verification/core/SamlDefinedIdentifiersVerifierSpec.kt
```diff
@@ -0,0 +1,281 @@
...
+            createResponse(
+                    identifierValue = maxLengthEntityId + "A",
+                    identifierFormat = ENTITY).let {
+                shouldThrow<SAMLComplianceException> {
+                    SamlDefinedIdentifiersVerifier(it).verify()
+                }.message?.shouldContain(SAMLCore_8_3_6_b.message)
+            }
+        }
+    }
+}
```
brjeter: ❗️ `verifyPersistentIdentifiers` and `verifyTransientIdentifiers` are not tested at all.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlDefinedIdentifiersVerifier.kt
```diff
@@ -148,89 +118,89 @@ internal class SamlDefinedIdentifiersVerifier(val node: Node) {
...
+                node.attributeNode(SP_NAME_QUALIFIER) != null ||
+                node.attributeNode("SPProvidedID") != null) {
             throw SAMLComplianceException.create(SAMLCore_8_3_6_a,
-                message = "No Subject element found.",
-                node = node)
+                    message = "Entity Identifier included a disallowed attribute.",
+                    node = node)
         }
-        node.nodeValue?.let {
+        node.textContent?.let {
```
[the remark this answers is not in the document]
bakejeyner: I'm not completely sure, however `textContent` is much better at getting the actual... text content from the node. Node value has returned a false null more times than I am comfortable with.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlDefinedIdentifiersVerifier.kt
bakejeyner: This was an SP test. The only thing with the "actions" is the `action` element, and the only thing with an `action` element is the `authzDecisionQuery` element, which is sent by the SP.

I'm not sure what the response to that query looks like; but it is probably just a response with assertions, and we aren't testing that protocol in the first place.

ctk/common/src/test/kotlin/org/codice/compilance/verification/core/SamlDefinedIdentifiersVerifierSpec.kt
bakejeyner: This is an awesome solution, I'm just worried about the information I would be losing doing it this way. It won't tell you why you failed; just that you failed on a certain email address.

brjeter: That seems fair to me 👍 I mean this is a unit test class, so reducing the amount of tests in here doesn't add just a ton of value.

ctk/common/src/test/kotlin/org/codice/compilance/verification/core/SamlDefinedIdentifiersVerifierSpec.kt
bakejeyner: Those verifications were merged in after my PR was up.

ctk/common/src/test/kotlin/org/codice/compilance/verification/core/SamlDefinedIdentifiersVerifierSpec.kt
bakejeyner: Implementing now.

Note that `NameFormat` cannot be null by Schema Validation, so that is not going to be implemented.

bakejeyner:
build now

PR #155, opened 2018-06-06
MAINT - Added unit test for CoreLogoutResponseProtocolVerifier

blen-desta:
Added unit test for `CoreLogoutResponseProtocolVerifier`

#### Checklist:

### DIFF
--- a/ctk/common/src/main/kotlin/org/codice/compliance/verification/core/responses/CoreLogoutResponseProtocolVerifier.kt
+++ b/ctk/common/src/main/kotlin/org/codice/compliance/verification/core/responses/CoreLogoutResponseProtocolVerifier.kt
@@ -26,7 +26,7 @@ import org.opensaml.saml.saml2.core.LogoutRequest
 import org.w3c.dom.Node
 
 class CoreLogoutResponseProtocolVerifier(logoutRequest: LogoutRequest, samlResponseDom: Node,
-    binding: SamlProtocol.Binding, private val expectedStatusCode: String? = null) :
+    binding: SamlProtocol.Binding, private val expectedSecondLevelStatusCode: String? = null) :
     ResponseVerifier(logoutRequest, samlResponseDom, binding) {
 
     override fun verify() {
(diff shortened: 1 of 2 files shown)


ctk/common/src/test/kotlin/org/codice/compilance/verification/core/response/CoreLogoutResponseProtocolVerifierSpec.kt
```diff
@@ -0,0 +1,100 @@
...
+                }.apply {
+                    this.message?.shouldContain(SAMLCore_3_7_3_2_b.message)
+                    this.message?.shouldContain(SAMLCore_3_7_3_2_d.message)
+                }
+            }
+        }
+    }
+
+    private fun createLogoutResponse(
+        secondStatusCodeValue: String = ""): String {
```
coyotesqrl: ❓ Are you anticipating future tests where the 2nd-level status code won't be provided to this fun? Your two current invocations pass in a value for this param.

blen-desta: That was what I was thinking but looking at it again, `CoreLogoutResponseProtocolVerifier` only validates second level status codes so I'll simplify it.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/responses/CoreLogoutResponseProtocolVerifier.kt
```diff
@@ -42,10 +42,11 @@ class CoreLogoutResponseProtocolVerifier(logoutRequest: LogoutRequest, samlRespo
                 ?.firstOrNull()
```
brjeter: This is out of scope of this PR, but I just had the thought of adding a `firstChild` extension function. That would simplify this logic significantly, and I wonder if it would simplify a lot of other places in the code as well.

ctk/common/src/test/kotlin/org/codice/compilance/verification/core/response/CoreLogoutResponseProtocolVerifierSpec.kt
```diff
@@ -0,0 +1,94 @@
...
+            Common.buildDom(createLogoutResponse(RESPONDER)).let {
+                shouldThrow<SAMLComplianceException> {
+                    CoreLogoutResponseProtocolVerifier(logoutRequest, it, HTTP_POST, SUCCESS)
+                        .verify()
+                }.apply {
+                    this.message?.shouldContain(SAMLCore_3_7_3_2_b.message)
+                    this.message?.shouldContain(SAMLCore_3_7_3_2_d.message)
+                }
+            }
+        }
```
brjeter: ❓ Should we also have a test for:
1. when there is no secondary status code (i.e. `null`) and we're expecting one, it will throw an exception (this could utilize the second test's code, just with different input)
2. When we're not expecting a status code, it can have any secondary status code and it won't throw an exception (I see some benefit in doing an array of a handful of values here to make sure none of them throw an exception)

blen-desta:
build now

brjeter:
Hero'd successfully ✅

PR #160, opened 2018-06-11
Disable SSO/SLO tests if those protocols aren't supported according to metadata

brjeter:
* Fixed so we aren't creating a new IdpMetadata object every time it's accessed
* Tagged all tests with SSO or SLO
* Added to the ProjectConfig to exclude tests that are tagged with a profile that's unsupported
#### Checklist:

### DIFF
--- a/ctk/idp/src/main/kotlin/io/kotlintest/provided/ProjectConfig.kt
+++ b/ctk/idp/src/main/kotlin/io/kotlintest/provided/ProjectConfig.kt
@@ -13,6 +13,7 @@
  */
 package io.kotlintest.provided
 
+import de.jupf.staticlog.Log
 import io.kotlintest.AbstractProjectConfig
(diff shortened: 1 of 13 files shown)


ctk/idp/src/main/kotlin/io/kotlintest/provided/ProjectConfig.kt
```diff
@@ -14,16 +14,32 @@
 package io.kotlintest.provided
```
coyotesqrl: ❓ Should have asked this a while back...does this need to live in this package?

brjeter: Yup. https://github.com/kotlintest/kotlintest/blob/master/doc/reference.md#project-config

PR #168, opened 2018-06-22
MAINT - Modified SimpleSign to accept more SigAlgs when validating

blen-desta:
Modified SimpleSign to accept more SigAlgs when validating.
Still uses `SHA1withRSA` or `SHA1withDSA` when signing

#### Checklist:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,6 +35,8 @@ project.ext {
     mockk_version = '1.7.15'
     junit_jupiter_version = '5.1.0'
     junit_platform_version = '1.1.1'
+    apache_wss4j_version='2.2.2'
+    apache_cxf_version='3.2.2'
 }
 
 allprojects {
--- a/ctk/common/src/main/java/org/codice/compliance/utils/sign/SimpleSign.java
+++ b/ctk/common/src/main/java/org/codice/compliance/utils/sign/SimpleSign.java
@@ -33,6 +33,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import org.apache.cxf.rs.security.saml.sso.SSOConstants;
+import org.apache.wss4j.common.WSS4JConstants;
 import org.apache.wss4j.common.crypto.CryptoType;
 import org.apache.wss4j.common.ext.WSSecurityException;
 import org.apache.wss4j.common.saml.OpenSAMLUtil;
@@ -44,6 +45,7 @@ import org.apache.wss4j.dom.saml.WSSSAMLKeyInfoProcessor;
 import org.apache.wss4j.dom.validate.Credential;
 import org.apache.wss4j.dom.validate.SignatureTrustValidator;
 import org.apache.wss4j.dom.validate.Validator;
+import org.apache.xml.security.algorithms.JCEMapper;
 import org.opensaml.saml.common.SAMLObjectContentReference;
 import org.opensaml.saml.common.SignableSAMLObject;
 import org.opensaml.saml.saml2.core.Assertion;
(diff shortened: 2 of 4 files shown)


build.gradle
```diff
@@ -35,6 +35,8 @@ project.ext {
     mockk_version = '1.7.15'
     junit_jupiter_version = '5.1.0'
     junit_platform_version = '1.1.1'
+    apache_wss4j_version='2.2.2'
+    apache_cxf_version='3.2.2'
```
coyotesqrl: ✏️ Should be CXF version 3.2.4 to match the version DDF is being upgraded to.

ctk/common/src/main/java/org/codice/compliance/utils/sign/SimpleSign.java
```diff
@@ -56,21 +58,20 @@
...
-
   private final SystemCrypto crypto;
 
-  private static final Map<String, String> URI_ALG_MAP = new HashMap<>();
-
-  static {
-    URI_ALG_MAP.put("http://www.w3.org/2000/09/xmldsig#dsa-sha1", "SHA1withDSA");
-    URI_ALG_MAP.put("http://www.w3.org/2000/09/xmldsig#rsa-sha1", "SHA1withRSA");
-  }
+  private static final ImmutableMap<String, String> URI_ALG_MAP =
```
coyotesqrl: ✏️ Should rename map or add a comment so the source of these mapped values isn't a mystery

ctk/common/src/main/java/org/codice/compliance/utils/sign/SimpleSign.java
```diff
@@ -56,21 +58,20 @@
...
+      ImmutableMap.<String, String>builder()
+          .put("http://www.w3.org/2001/04/xmldsig-more#rsa-md5", "md5WithRSA")
+          .put(WSS4JConstants.RSA, "sha1WithRSA")
+          .put(WSS4JConstants.RSA_SHA256, "sha256WithRSA")
+          .put("http://www.w3.org/2001/04/xmldsig-more#rsa-sha384", "sha384WithRSA")
+          .put("http://www.w3.org/2001/04/xmldsig-more#rsa-sha512", "sha512WithRSA")
+          .put(WSS4JConstants.DSA, "sha1WithDSA")
+          .build();
 
   public SimpleSign() throws IOException {
```
coyotesqrl: ❓ Wasn't this class much closer - identical(?) - to the DDF version previously? If it was and how now diverged, we might want to document the ways in which it is written and operates that are different from DDF.

ctk/common/src/main/java/org/codice/compliance/utils/sign/SimpleSign.java
```diff
@@ -327,36 +327,22 @@ private void validateSignatureAndSamlKey(Signature signature, SAMLKeyInfo samlKe
...
   private java.security.Signature getSignature(X509Certificate certificate, PrivateKey privateKey)
       throws SignatureException {
-    String jceSigAlgo = "SHA1withRSA";
-    if ("DSA".equalsIgnoreCase(certificate.getPublicKey().getAlgorithm())) {
-      jceSigAlgo = "SHA1withDSA";
-    }
-
     java.security.Signature signature;
     try {
+      String jceSigAlgo = String.format("sha1with%s", certificate.getPublicKey().getAlgorithm());
```
brjeter: ❓ In @coyotesqrl's DDF PR, he just calls `certificate.getSigAlgName();` instead of this line. Did you test with that? Did that include the `sha` piece of the name?

bakejeyner: `certificate.getSigAlgName()` returns the algorithm that the CA used to sign the certificate, not the algorithm of the public key within the certificate.

brjeter: Did we decide that was incorrect? I don't remember.

bakejeyner: That was my understanding 🤷‍♂️. @blen-desta do you concur?

blen-desta: Yes, `certificate.getSigAlgName()` returns the alg used by the CA

brjeter: So this is the main thing that needs to be fixed in the DDF PR. @coyotesqrl do you want one of us to comment on the DDF PR with this context?

coyotesqrl: Comment, tell me what to do, paste a patch...you know, whatever _you_ guys think you should do.

ctk/common/src/main/java/org/codice/compliance/utils/sign/SimpleSign.java
brjeter: ❗️ We should get rid of this custom map in place of just using https://santuario.apache.org/Java/api/org/apache/xml/security/algorithms/JCEMapper.html. You can call `JCEMapper.translateURItoJCEID(sigAlg);` where `sigAlg` is the URI format and it'll spit out either `null` or the JCE format i.e. "SHA1withDSA".

blen-desta:
build now

PR #169, opened 2018-07-06
MAINT - Fixing test issues that were discovered against Keycloak

brjeter:
- Removed Trust validation from SimpleSign since we can't expect any third-party IdP to put its cert in our keystore
- Preprocess the authn response in SLO tests, so that it's decrypted for the `SubjectComparisonVerifier`
- Added TestListener that always sets back to the default SP

#### Checklist:


ctk/common/src/main/java/org/codice/compliance/utils/sign/SimpleSign.java
```diff
@@ -279,17 +276,6 @@ public void validateSignature(Signature signature) throws SignatureException {
     }
 
     validateSignatureAndSamlKey(signature, samlKeyInfo);
-
```
bakejeyner: ❓ Why is this being taken out?

brjeter: It's being taken out because otherwise it requires users to put their IdP's cert into our keystore which is painful and unnecessary. It's unnecessary because this is a testing tool that should only be run in a test environment. Trusting the certs that were used for signing is significantly less important in such a scenario.

PR #174, opened 2018-07-10
DDF-3893 Add NameIDPolicy Format support

bakejeyner:
Un-comments out the sections ingored because of NameIDPolicy.
Do not merge until [this DDF PR](url) is merged.

#### Checklist:


ctk/idp/src/main/kotlin/org/codice/compliance/web/sso/PostSSOTest.kt
```diff
@@ -109,8 +109,7 @@ class PostSSOTest : StringSpec() {
 
         // TODO When DDF is fixed to return NameID format based on NameIDPolicy,
```
coyotesqrl: ❓ Should this comment be removed as well?

library/src/main/kotlin/org/codice/compliance/Common.kt
```diff
@@ -32,7 +32,8 @@ import javax.xml.transform.TransformerFactory
 import kotlin.test.currentStackTrace
 
 const val IMPLEMENTATION_PATH = "implementation.path"
-const val DEFAULT_IMPLEMENTATION_PATH = "implementations/ddf"
+@Suppress("TopLevelPropertyNaming")
+val DEFAULT_IMPLEMENTATION_PATH = "implementations${File.separator}ddf"
```
brjeter: ❓ Does this enable you to run the script from the command-line on Windows? That's something we should probably fix pretty soon.

bakejeyner: It fixes something windows related. I think it was even in the code things were messing up (some tests were failing because they couldn't find a file because of that separator).

bakejeyner:
build now

PR #177, opened 2018-07-17
MAINT - Added a DDF profile flag to allow the profile to be run against non-DDF IDPs

blen-desta:
Added a DDF profile flag to allow the profile to be run against non-DDF IDPs.

The point of the DDF profile is for it to be run against non-DDF IdP to see if they can interoperate successfully. 
> Change for dev - add the -ddf flag when running the tests

#### Checklist:


deployment/distribution/src/main/kotlin/org/codice/ctk/Command.kt
```diff
@@ -96,6 +81,32 @@ fun main(args: Array<String>) {
     launchTests()
 }
 
+private fun createParser(): Parser {
+    val parser = Parser()
```
coyotesqrl: ✏️ `return Parser().apply { ...`

blen-desta:
build now

PR #180, opened 2018-07-17
SLO - Added tests for using a DSA SHA256 signature for DDF profile

brjeter:
- Updated `SimpleSign` to be more similar to DDF's version
- Commented on every difference between our version of `SimpleSign` and DDF's version.
- Added a test for Redirect/POST SSO and Redirect/POST SLO which signs the request with SHA256withDSA
#### Checklist:

### DIFF
--- a/ctk/common/src/main/java/org/codice/compliance/utils/sign/SimpleSign.java
+++ b/ctk/common/src/main/java/org/codice/compliance/utils/sign/SimpleSign.java
@@ -56,7 +56,7 @@ import org.opensaml.xmlsec.signature.support.SignatureConstants;
 import org.opensaml.xmlsec.signature.support.SignatureValidator;
 import org.opensaml.xmlsec.signature.support.provider.ApacheSantuarioSignatureValidationProviderImpl;
 
-// NOTE: "DDF DIFFERENCE" means something that is different in the DDF version of this class.
+// NOTE: "DDF Difference" means something that is different in the DDF version of this class.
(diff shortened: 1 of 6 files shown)


ctk/common/src/main/java/org/codice/compliance/utils/sign/SimpleSign.java
```diff
@@ -56,16 +56,53 @@
...
   public SimpleSign() throws IOException {
     crypto = new SystemCrypto(getCurrentSPHostname());
+    rsaAlgoUri = WSS4JConstants.RSA;
+    rsaAlgoJce = JCEMapper.translateURItoJCEID(rsaAlgoUri);
+    dsaAlgoUri = WSS4JConstants.DSA;
+    dsaAlgoJce = JCEMapper.translateURItoJCEID(dsaAlgoUri);
+  }
+
+  // DDF Difference: The signing algorithm is not configurable in DDF.
+  public SimpleSign(String rsaAlgoUri, String dsaAlgoUri) throws IOException {
```
blen-desta: ❓ Should we just pass in one algorithm instead of two? Cause we'll just be using the one

brjeter:
@bakejeyner We should only be testing SHA256 with DSA. It's sufficient to test only SHA1 with RSA. https://github.com/connexta/saml-conformance/pull/180#discussion_r203210252

brjeter:
build now

brjeter:
build now

brjeter:
build now

bantillo:
manual hero verified, works if ddf is installed using profile:install standard

PR #184, opened 2018-07-23
MAINT - Added a clean display of all tests run in the console.

brjeter:
Example output:
screen shot 2018-07-23 at 3 41 26 pm

#### Checklist:


deployment/distribution/src/main/kotlin/org/codice/ctk/Command.kt
```diff
@@ -110,18 +115,48 @@ private fun launchTests() {
...
+        System.out.apply {
+            println()
+            println("----------------------------------")
+            println("SAML Conformance Test Kit Starting")
+            println("----------------------------------")
+        }
+    }
+
+    override fun executionFinished(testIdentifier: TestIdentifier?,
+                                   testExecutionResult: TestExecutionResult?) {
```
blen-desta: ❓ Is it possible to change `testIdentifier` and `testExecutionResult` to `TestIdentifier` and `TestExecutionResult` without the `?`? That way you don't need the `?.let` or the null check

brjeter: I was, in fact, able to do that! 🎉 I didn't know I could do that with a function override.

PR #185, opened 2018-07-27
Updates build files to kotlin dsl and updates license headers

coyotesqrl:
#### Checklist:


deployment/distribution/build.gradle.kts
```diff
@@ -0,0 +1,73 @@
...
+    compile("us.jimschubert:kopper-typed:0.0.3")
+    compile("org.fusesource.jansi:jansi:1.17.1")
+}
+
+// We don"t want to include the implementation jars in the classpath
+// since they"re loaded from a specific directory.
+tasks {
+    "copyToLib"(Copy::class) {
+        from(configurations.runtime) {
+            // TODO RAP 27 Jul 18: Fix this exclusion
```
coyotesqrl: @blen-desta I was hoping you could specifically take a look at this and see if you can help me figure out if we still need it and what it should look like.

coyotesqrl:
build now

detekt.yml
```diff
@@ -270,6 +270,9 @@ potential-bugs:
 
 style:
   active: true
+#  TODO This should possibly be enabled; it should be discussed
```
coyotesqrl: ❓ I got a ton of failures for this check that we hadn't been getting before. I don't know why it changed.

We need to decide if we want to enforce braces on one-line ifs or not. It feels a little un-kotliny, but it is a tighter match with our Java standard so there are arguments both ways.

coyotesqrl:
build now

coyotesqrl:
build now

deployment/distribution/build.gradle.kts
brjeter: I would expect this to still be necessary. You should be able to manually verify that by removing the jar from the `implementations/ddf` folder under the exploded distro and running the default options. The tests would still run because the ddf jar is on the classpath, even though it's not being loaded from the implementation folder.

deployment/distribution/build.gradle.kts
```diff
@@ -0,0 +1,80 @@
...
+dependencies {
+    compile(project(":external:implementations:samlconf-ddf-impl"))
+    compile(project(":external:implementations:samlconf-keycloak-impl"))
+    compile(project(":library"))
+    compile(project(":external:samlconf-plugins-api"))
+    compile(project(":ctk:common"))
+    compile(project(":ctk:idp"))
+
+    compile("us.jimschubert:kopper-typed:0.0.3")
+    compile("org.fusesource.jansi:jansi:1.17.1")
```
brjeter: ❓ Do we want to centralize all dependency versions into the `Versions` object, even single usages like this?

library/build.gradle.kts
```diff
@@ -0,0 +1,43 @@
...
+
+plugins {
+    `maven-publish`
+}
+
+dependencies {
+    compile(Libs.cxfSsoSaml)
+    compile("com.google.http-client:google-http-client:1.22.0")
+    compile("org.keyczar:keyczar:0.66")
+    compile("net.sf.jtidy:jtidy:r938")
```
brjeter: ❓ Same question about putting these versions in the central `Versions` object.

deployment/distribution/build.gradle.kts
coyotesqrl: Certainly can. I hadn't done the one-offs, but it's not a big deal to do so.

build.gradle.kts
```diff
@@ -0,0 +1,131 @@
...
+}
+
+val compileKotlin: KotlinCompile by tasks
+compileKotlin.kotlinOptions {
+    jvmTarget = Versions.javaTarget
+}
+val compileTestKotlin: KotlinCompile by tasks
+compileTestKotlin.kotlinOptions {
+    jvmTarget = Versions.javaTarget
+}
```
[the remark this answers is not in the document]
coyotesqrl: We had it. It was in a different location in the file, though.

deployment/distribution/build.gradle.kts
coyotesqrl: @blen-desta Nope. That's not supported. That's why I was hoping for a little help. ;)

deployment/distribution/build.gradle.kts
```diff
@@ -0,0 +1,70 @@
...
+
+    compile(Libs.kopperTyped)
+    compile(Libs.jansi)
+}
+
+// We don"t want to include the implementation jars in the classpath
+// since they"re loaded from a specific directory.
+tasks {
+    "copyToLib"(Copy::class) {
+        from(configurations.runtime.exclude(group = "org.codice.samlconf.implementations"))
```
coyotesqrl: This should be correct now. It does exclude the implementation jars and, transitively, `metainf-services-1.1.jar` and `gson-2.8.0.jar`. I did notice that because of that latter transitive exclusion that an older version of the gson jar - `gson-1.4.jar` - gets pulled into the lib directory.

This is the same behavior as on master.

blen-desta:
Checking if build is fixed
build now

blen-desta:
CI build is still broken.

PR #198, opened 2018-08-22
Disabled email name id verification

blen-desta:
### Description of the Change
Disabled email name id verification because it isn't a MUST.

```
CORE 8.3.2 Email Address
URI: urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress
Indicates that the content of the element is in the form of an email address, 
specifically "addr-spec" as defined in IETF RFC 2822 Section 3.4.1. An addr-spec 
has the form local-part@domain. Note that an addr-spec has no phrase (such as a 
common name) before it, has no comment (text surrounded in parentheses) after it, 
and is not surrounded by "<" and ">".
```

### Verification Process
Successful CI build

### Checklist:

### DIFF
--- a/ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlDefinedIdentifiersVerifier.kt
+++ b/ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlDefinedIdentifiersVerifier.kt
@@ -6,6 +6,7 @@ http://www.gnu.org/licenses/lgpl.html
 */
 package org.codice.compliance.verification.core
 
+import org.codice.compliance.Common.Companion.idpMetadataObject
 import org.codice.compliance.SAMLComplianceException
 import org.codice.compliance.SAMLCore_8_2_2_a
 import org.codice.compliance.SAMLCore_8_2_3_a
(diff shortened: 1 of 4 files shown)


ctk/common/src/test/kotlin/org/codice/compilance/verification/core/SamlDefinedIdentifiersVerifierSpec.kt
```diff
@@ -130,7 +129,7 @@ class SamlDefinedIdentifiersVerifierSpec : StringSpec() {
             }
         }
 
-        /* 8.3.2 Email Address */
+        /* 8.3.2 Email Address
         "valid Email name identifier" {
```
coyotesqrl: ❓ Why commented out and not enabled=false?

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SamlDefinedIdentifiersVerifier.kt
```diff
@@ -64,7 +64,7 @@ internal class SamlDefinedIdentifiersVerifier(val node: Node) {
     /** 8 SAML-Defined Identifiers */
     fun verify() {
         verifyAttributeNameFormatIdentifiers()
-        verifyEmailAddressIdentifier()
+        // verifyEmailAddressIdentifier()
```
coyotesqrl: ✏️ Should leave a comment here indicating the reason for the commented out verification step and what it would require to re-enable it.

blen-desta:
build now

PR #200, opened 2018-09-13
Modify invalid subject tests to not expect an error immediately

bakejeyner:
### Description of the Change

Before, the invalid subject error tests use to expect the IdP to error out directly after sending the `AuthnRequest`. Now the tests will send the `AuthnRequest`, attempt the user login, then expect the error.

### Verification Process

The invalid subject error tests should pass when running the CTK against [this branch of DDF](https://github.com/codice/ddf/pull/3733).

### Checklist:

### DIFF
--- a/ctk/idp/src/main/kotlin/org/codice/compliance/web/sso/error/PostSSOErrorTest.kt
+++ b/ctk/idp/src/main/kotlin/org/codice/compliance/web/sso/error/PostSSOErrorTest.kt
@@ -16,12 +16,14 @@ import org.codice.compliance.SAMLComplianceException
 import org.codice.compliance.SAMLCore_3_2_1_e
 import org.codice.compliance.SAMLProfiles_4_1_4_1_a
 import org.codice.compliance.SAMLProfiles_4_1_4_1_b
+import org.codice.compliance.saml.plugin.IdpSSOResponder
 import org.codice.compliance.utils.EXAMPLE_RELAY_STATE
 import org.codice.compliance.utils.INCORRECT_DESTINATION
 import org.codice.compliance.utils.RELAY_STATE_GREATER_THAN_80_BYTES
 import org.codice.compliance.utils.REQUESTER
 import org.codice.compliance.utils.SSOCommon.Companion.createDefaultAuthnRequest
 import org.codice.compliance.utils.SSOCommon.Companion.sendPostAuthnRequest
+import org.codice.compliance.utils.TestCommon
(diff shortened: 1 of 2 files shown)


coyotesqrl:
Is there an issue for this PR?

ctk/idp/src/main/kotlin/org/codice/compliance/web/sso/error/PostSSOErrorTest.kt
```diff
@@ -93,9 +95,16 @@ class PostSSOErrorTest : StringSpec() {
                     val encodedRequest = signAndEncodePostRequestToString(authnRequest,
                         EXAMPLE_RELAY_STATE)
                     val response = sendPostAuthnRequest(encodedRequest)
+                    BindingVerifier.verifyHttpStatusCode(response.statusCode)
 
-                    if (!isLenient || !BindingVerifier.isErrorHttpStatusCode(response.statusCode)) {
-                        val samlResponseDom = response.getBindingVerifier().decodeAndVerifyError()
+                    val finalHttpResponse =
+                            TestCommon.getImplementation(IdpSSOResponder::class)
+                                    .getResponseForPostRequest(response)
```
brjeter: I think I'm okay with this change until we run into an IdP that acts differently, but I think our IdP needing more interaction before validating the Subject is implementation-dependent.

bakejeyner: I agree that this is implementation dependent. I also agree that in order to fix this issue, a separate effort should be made to look into:
1. How many of our tests are actually implementation dependent
2. The best solution to solve this implementation dependence.

Implementation dependent example: IdP1 could take the AuthnRequest Subject and check it against its user database, sending an error response without having to log any user in. Whereas IdP2 (lookin' at you DDF) may log the user in, then compare that user to the Subject sent in the AuthnRequest.

bakejeyner:
@coyotesqrl There was no issue at the time of your comment, however I have created one and made links from the PR to the issue and vice versa.

bakejeyner:
build now

blen-desta:
test ddf

blen-desta:
Oh forgot that https://github.com/codice/ddf/pull/3733 needs to be merged for that test to pass.
build now

brjeter:
Hero successful ✅

blen-desta:
build now

PR #206, opened 2018-11-29
Improve plugin section of readme

bakejeyner:
### Description of the Change

Improving the plugin jar section of the root readme.

N/A

### Verification Process

N/A

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,20 +103,12 @@ This means that the CTK doesn't know how to give the IdP the principal's credent
 Therefore each IdP must write a plugin which will process the authentication of a principal.
 See [4.1 Web Browser SSO Profile](https://www.oasis-open.org/committees/download.php/56782/sstc-saml-profiles-errata-2.0-wd-07.pdf) for more information on the different steps for the SSO profile.
 
-###### Examining the DDF plugin jar
-In the case of DDF's IdP plugin:
-
-1. The plugin implements the [IdpSSOResponder](external/samlconf-plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpSSOResponder.java) interface.
-1. The plugin receives the login page from DDF's IdP as an http response in the interface methods.
-1. The plugin parses everything it needs from the login page and gets ready to send a request back to DDF's IdP.
-1. The plugin sends an http request to the proper location, with the principal's credentials, in the format that DDF's IdP expects the resulting request.
-
 ###### How to implement a plugin jar?
 
-1. Implement a plugin jar for the IdP's authentication implementation (see "Examining the DDF plugin jar" and step 4 in "Why is a plugin required?"). \
+1. Implement a plugin jar for the IdP's authentication implementation. \
 Write a Java or Kotlin class that implements the [IdpSSOResponder](external/samlconf-plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpSSOResponder.java) interface. \
 [DDFIdpSSOResponderProvider](external/implementations/samlconf-ddf-impl/src/main/kotlin/org/codice/compliance/saml/plugin/ddf/DDFIdpSSOResponderProvider.kt) and
-[KeycloakIdpSSOResponderProvider](external/implementations/samlconf-keycloak-impl/src/main/kotlin/org/codice/compliance/saml/plugin/keycloak/KeycloakIdpSSOResponderProvider.kt) can be used as examples. \
+[KeycloakIdpSSOResponderProvider](external/implementations/samlconf-keycloak-impl/src/main/kotlin/org/codice/compliance/saml/plugin/keycloak/KeycloakIdpSSOResponderProvider.kt) can be used as examples (see [Examining the DDF plugin jar](#examining-the-ddf-plugin-jar)). \
 **NOTE**: The implementation class is implementing a Java service therefore either
    * the `@MetaInfServices` annotation should be used (which uses the kapt plugin) or
    * a hand-crafted services reference file `META-INF/services/org.codice.compliance.saml.plugin.IdpSSOResponder` will need to be added to the jar (see below)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -99,12 +99,21 @@ When logging in to an IdP using the Single Sign-On protocol, there are 6 steps:
...
-Since the authentication step is implementation specific, each IdP must write a plugin which will process the authentication of a principal.
+This means that the CTK doesn't know how to give the IdP the principal's credentials, since the authentication step is implementation specific. \
+Therefore each IdP must write a plugin which will process the authentication of a principal.
 See [4.1 Web Browser SSO Profile](https://www.oasis-open.org/committees/download.php/56782/sstc-saml-profiles-errata-2.0-wd-07.pdf) for more information on the different steps for the SSO profile.
 
+###### Examining the DDF plugin jar
+In the case of DDF's IdP plugin:
+
+1. The plugin implements the [IdpSSOResponder](external/samlconf-plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpSSOResponder.java) interface.
+1. The plugin receives the login page from DDF's IdP as an http response in the interface methods.
```
coyotesqrl: ✏️ I know that this will work fine in Markdown, but it makes me itchy for the numbers not to be sequential in the source.

blen-desta: I think we should be consistent throughout the README so if you change this, you should probably change the others to increment too

blen-desta: ✏️ ".. as an HTTP response ..."

README.md
```diff
@@ -99,12 +99,21 @@ When logging in to an IdP using the Single Sign-On protocol, there are 6 steps:
...
+1. The plugin parses everything it needs from the login page and gets ready to send a request back to DDF's IdP.
+1. The plugin sends an http request to the proper location, with the principal's credentials, in the format that DDF's IdP expects the resulting request.
+
```
brjeter: ❓ Would it be helpful for us to put a note with a link to the source code of the DDF plugin? My flinch reaction is no because it's too low-level, but the audience for this section is developers so it wouldn't be too low-level.

README.md
```diff
@@ -99,12 +99,21 @@ When logging in to an IdP using the Single Sign-On protocol, there are 6 steps:
...
 ###### How to implement a plugin jar?
 
-1. Implement a plugin jar for the IdP's authentication implementation. \
+1. Implement a plugin jar for the IdP's authentication implementation (see "Examining the DDF plugin jar" and step 4 in "Why is a plugin required?"). \
```
brjeter: ✏️ You could turn the "Examining the DDF plugin jar" and "Why is a plugin required?" phrases into links.

README.md
```diff
@@ -99,12 +99,21 @@ When logging in to an IdP using the Single Sign-On protocol, there are 6 steps:
 1. Service Provider grants or denies access to Principal
 
 The specification mentions that in step 4, "the principal is identified by the identity provider by some means outside the scope of \[the Single Sign-On\] profile".
-Since the authentication step is implementation specific, each IdP must write a plugin which will process the authentication of a principal.
+This means that the CTK doesn't know how to give the IdP the principal's credentials, since the authentication step is implementation specific. \
+Therefore each IdP must write a plugin which will process the authentication of a principal.
 See [4.1 Web Browser SSO Profile](https://www.oasis-open.org/committees/download.php/56782/sstc-saml-profiles-errata-2.0-wd-07.pdf) for more information on the different steps for the SSO profile.
 
+###### Examining the DDF plugin jar
```
blen-desta: ✏️ I think this should go under `How to implement a plugin jar` cause it talks about a specific plugin. I think we should do:
- Why is a plugin required?
- How to implement a plugin jar?
- Examining the DDF plugin jar

README.md
```diff
@@ -99,12 +99,21 @@ When logging in to an IdP using the Single Sign-On protocol, there are 6 steps:
...
+In the case of DDF's IdP plugin:
+
+1. The plugin implements the [IdpSSOResponder](external/samlconf-plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpSSOResponder.java) interface.
+1. The plugin receives the login page from DDF's IdP as an http response in the interface methods.
+1. The plugin parses everything it needs from the login page and gets ready to send a request back to DDF's IdP.
```
blen-desta: ✏️ "... sends an HTTP request ..."

README.md
blen-desta: ✏️ Step 4 in `Why is a plugin required` just says "Identity Provider identifies Principal" so I'm not sure why we need the see step 4 and for the `see "Examining the DDF plugin jar"` part it would make more sense to put it one sentence after. Cause we do talk about the plugins. So for example something like:
```
Implement a plugin jar for the IdP's authentication implementation.
Write a Java or Kotlin class that implements the IdpSSOResponder interface. 
DDFIdpSSOResponderProvider and KeycloakIdpSSOResponderProvider can be used as examples. 
See [Examining the DDF plugin jar](#examining-the-ddf-plugin-jar) for a step 
by step description of what the DDF plugin is doing.
```

README.md
bakejeyner: I actually prefer not sequentially numbering the list in the source. That way if we need to add a step, we don't have to refactor all of the numbers.

README.md
bakejeyner: I prefer links 👍

bakejeyner:
build now

README.md
```diff
@@ -103,20 +103,12 @@ This means that the CTK doesn't know how to give the IdP the principal's credent
...
 See [4.1 Web Browser SSO Profile](https://www.oasis-open.org/committees/download.php/56782/sstc-saml-profiles-errata-2.0-wd-07.pdf) for more information on the different steps for the SSO profile.
 
-###### Examining the DDF plugin jar
-In the case of DDF's IdP plugin:
-
-1. The plugin implements the [IdpSSOResponder](external/samlconf-plugins-api/src/main/java/org/codice/compliance/saml/plugin/IdpSSOResponder.java) interface.
-1. The plugin receives the login page from DDF's IdP as an http response in the interface methods.
-1. The plugin parses everything it needs from the login page and gets ready to send a request back to DDF's IdP.
-1. The plugin sends an http request to the proper location, with the principal's credentials, in the format that DDF's IdP expects the resulting request.
-
```
bakejeyner: This section was moved to line 121.

PR #211, opened 2018-12-19
MAINT - Improved Reporting for CORE

blen-desta:
### Description of the Change
- Added the ability to have a report and continue running tests when an error occurs
- Used a Report object (singleton) to add errors instead of throwing exceptions
- Added the ability to run quietly which is just the report without the errors
- Added a codes field in SAMLComplianceException and the ability to compare them

**Hopefully, the separate commits help.**
Updates:
- I have location-wise moved the report package to be near the SAMLComplainceException one
- Now both a SAMLRefMessage and a SAMLComplainceException have knowledge of what section the exception corresponds to so no need of passing section around.

N/A

### Verification Process
- Make sure that failures are reported successfully 
- Make sure tests are skipped successfully 
- Make sure DDF passes

## **NOT** quiet mode:
screen shot 2019-01-03 at 10 43 10 am
screen shot 2019-01-03 at 10 43 25 am

## Quiet mode:
screen shot 2019-01-03 at 11 20 52 am
screen shot 2019-01-03 at 11 23 22 am

### Checklist:


ctk/common/src/main/kotlin/org/codice/compliance/utils/report/Report.kt
```diff
@@ -0,0 +1,141 @@
...
+import org.fusesource.jansi.Ansi
+import org.junit.platform.engine.TestExecutionResult.Status.FAILED
+import org.junit.platform.engine.TestExecutionResult.Status.SUCCESSFUL
+
+object Report {
+
+    private val errorMessages = mutableMapOf<Sections, MutableList<SAMLComplianceException>?>()
+    private val coreHeaderSize = Sections.CORE_1.name.length
+
+    init {
```
coyotesqrl: ✏️ Alternately, could do this in the declaration with
```kotlin
    private val errorMessages = Sections.values().map {
        it to mutableListOf<SAMLComplianceException?>()
    }.toMap().toMutableMap()
```

ctk/common/src/main/kotlin/org/codice/compliance/utils/report/Report.kt
```diff
@@ -0,0 +1,141 @@
...
+        Sections.values().forEach { errorMessages[it] = mutableListOf() }
+    }
+
+    private val reportQuietly by lazy {
+        System.getProperty(QUIET_REPORT) == "true"
+    }
+
+    fun addErrorMessage(section: Sections, error: SAMLComplianceException): Report {
+        // Error already added
+        if (errorMessages[section]?.isNotEmpty() == true &&
```
coyotesqrl: ✏️ The second test should be sufficient. I'm not sure there's enough of a performance improvement here to warrant the compound condition.

On second look, does it really matter at all if the error is added to the section multiple times? If you convert it from a mutableList to a mutableSet, at least?

ctk/common/src/main/kotlin/org/codice/compliance/utils/report/Report.kt
```diff
@@ -0,0 +1,141 @@
...
+        errorMessages[section]?.add(error)
+        return this
+    }
+
+    fun getExceptionMessages(section: Sections): String? {
+        return errorMessages[section]?.joinToString()
+    }
+
+    fun hasErrors(): Boolean {
+        errorMessages.forEach {
```
coyotesqrl: ✏️ This should be equivalent:
```kotlin
return errorMessages.values.any { it.isNotEmpty() }
```

ctk/common/src/main/kotlin/org/codice/compliance/utils/report/Report.kt
```diff
@@ -0,0 +1,141 @@
...
+                print(it)
+            }
+
+            // CORE
+            if (it.name.startsWith("CORE")) {
+                if (it == Sections.CORE_1) {
+                    println("CORE")
+                    println("\t${it.title}")
+                } else {
+                    if (it.name.length == coreHeaderSize) {
```
coyotesqrl: ❗️ This is really brittle. I think it would be better to enhance the enum so it carried some meta information about itself, info like `isTopLevel` and such. I've posted a gist with that and the rest of my suggested changes for this class here - https://gist.github.com/coyotesqrl/0a642eb17b4ab89cbe7e24a9d0277189

ctk/common/src/main/kotlin/org/codice/compliance/utils/report/Report.kt
blen-desta: Not if it’s a set

ctk/common/src/main/kotlin/org/codice/compliance/utils/report/Report.kt
blen-desta: Oh I see I can add multiple values to it 👌

deployment/distribution/src/main/kotlin/org/codice/ctk/TestRunner.kt
```diff
@@ -49,51 +45,23 @@ internal class TestRunner {
...
-                            ${getResultDisplay(testExecutionResult.status)}""")
-        }
-
-        private fun getResultDisplay(status: TestExecutionResult.Status): Any {
-            return when (status) {
-                TestExecutionResult.Status.SUCCESSFUL ->
-                    Ansi.ansi().fgGreen().a(status.name).reset()
-                TestExecutionResult.Status.FAILED ->
-                    Ansi.ansi().fgRed().a(status.name).reset()
-                else -> status.name
```
brjeter: In order to preserve the information on the particular tests that failed, you could either:
1. Keep this `TestNameListener`, in addition to printing the `Report` at the end.
2. Somehow make either the `Report` or the `SAMLComplianceException`s aware of the current test being run.

Number 2 sounds really messy code-wise. There is an advantage to number 1, which is that it will show progress to the user as the tests are running. With this change as it is now, the user will not see any progress; they just get a summary at the end.

brjeter: And I'm realizing now that even that won't work as it is because the tests won't necessarily fail anymore... What if we gave the report some notion of "staged exceptions". In addition to preserving the exceptions it gets, it keeps a cache that the `TestNameListener` would grab at the end of the test and then clear out for the next test.

ctk/common/src/test/kotlin/org/codice/compilance/utils/report/ReportSpec.kt
```diff
@@ -0,0 +1,29 @@
...
+
+        "test adding an error to the report" {
+            Report.addErrorMessage(GENERAL,
+                    SAMLComplianceException.create(SAMLGeneral_a, message = "message"))
+        }
+
+        "test adding errors from the same section in the report" {
+        }
+
+        "test getting the report" {
```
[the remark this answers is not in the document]
blen-desta: Yes 😆

deployment/distribution/src/main/kotlin/org/codice/ctk/TestRunner.kt
```diff
@@ -73,26 +71,25 @@ internal class TestRunner {
...
             testIdentifier: TestIdentifier,
             testExecutionResult: TestExecutionResult
         ) {
-            if (testIdentifier.isTest)
-                System.out.println(
-                        """${testIdentifier.displayName}
-                            ${getResultDisplay(testExecutionResult.status)}""")
-        }
+            if (testIdentifier.isTest) {
+                System.out.print(testIdentifier.displayName)
```
coyotesqrl: ✏️ Should be able to just use `print` here

library/src/main/kotlin/org/codice/compliance/SAMLComplianceException.kt
```diff
@@ -10,36 +10,56 @@ import de.jupf.staticlog.Log
 import de.jupf.staticlog.core.LogLevel
 import org.w3c.dom.Node
 
+@Suppress("StringLiteralDuplication")
 class SAMLComplianceException : Exception {
 
-    private constructor(message: String) : super(message)
+    val errorCodes = mutableListOf<SAMLSpecRefMessage>()
```
coyotesqrl: ✏️ Instead of making this mutable, could do the following:
```kotlin
    val errorCodes: List<SAMLSpecRefMessage>

    private constructor(codes: Array<out SAMLSpecRefMessage>, message: String) : super(message) {
        errorCodes = listOf(*codes)
    }

    private constructor(codes: Collection<SAMLSpecRefMessage>, message: String) : super(message) {
        errorCodes = codes.toList()
    }

    private constructor(
        codes: Array<out SAMLSpecRefMessage>,
        message: String,
        cause: Throwable
    ) : super(message, cause) {
        errorCodes = listOf(*codes)
    }

    private constructor(
        codes: Collection<SAMLSpecRefMessage>,
        message: String,
        cause: Throwable
    ) : super(message, cause) {
        errorCodes = codes.toList()
    }
```

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/NameIDPolicyVerifier.kt
```diff
@@ -46,22 +46,23 @@ class NameIDPolicyVerifier(private val samlResponseDom: Node, private val policy
...
             if (spnq != spNameQualifier) {
-                throw SAMLComplianceException.create(SAMLCore_3_4_1_1_b,
-                        message = "A NameID element was found with a SPNameQualifier " +
-                                "attribute value of $spnq instead of " +
-                                "$spNameQualifier.",
-                        node = nameId)
+                Report.addExceptionMessage(CORE_3_4,
+                        SAMLComplianceException.create(SAMLCore_3_4_1_1_b,
+                                message = """A NameID element was found with a SPNameQualifier
+                                        attribute value of $spnq instead of $spNameQualifier.""",
```
brjeter: ❓ Maybe I'm remembering incorrectly but isn't changing this to a raw string going to include the two lines, with a newline in between, in the output message?

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/NameIDPolicyVerifier.kt
```diff
@@ -72,17 +73,19 @@ class NameIDPolicyVerifier(private val samlResponseDom: Node, private val policy
...
 
             if (subjects.any { it.children("EncryptedID").isEmpty() }) {
-                throw SAMLComplianceException.create(SAMLCore_3_4_1_1_a,
-                        message = "An Assertion element was found without an EncryptedID element" +
-                                " in its Subject element.",
-                        node = samlResponseDom)
+                Report.addExceptionMessage(CORE_3_4,
+                        SAMLComplianceException.create(SAMLCore_3_4_1_1_a,
+                                message = """An Assertion element was found without an EncryptedID
+                                    element in its Subject element.""",
```
brjeter: Same comment about raw strings.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/NameIDPolicyVerifier.kt
```diff
@@ -72,17 +73,19 @@ class NameIDPolicyVerifier(private val samlResponseDom: Node, private val policy
...
 
             if (subjects.any { it.children("NameID").isNotEmpty() }) {
-                throw SAMLComplianceException.create(SAMLCore_3_4_1_1_a,
-                        message = "An Assertion element was found with a NameID element in " +
-                                "its Subject element.",
-                        node = samlResponseDom)
+                Report.addExceptionMessage(CORE_3_4,
+                        SAMLComplianceException.create(SAMLCore_3_4_1_1_a,
+                                message = """An Assertion element was found with a NameID element
+                                    in its Subject element.""",
```
brjeter: Same comment about raw strings.

ctk/common/src/test/kotlin/org/codice/compilance/ReportListener.kt
```diff
@@ -0,0 +1,21 @@
...
+import io.kotlintest.Description
+import io.kotlintest.extensions.TestListener
+import org.codice.compliance.utils.report.Report
+
+/**
+ * Listener used to reset the Report map for testing purposes before each test
+ */
+object ReportListener : TestListener {
+
+    override fun beforeTest(description: Description) {
```
brjeter: ✏️ The first test wouldn't need this beforehand. Technically, you could avoid the extra reset by doing `afterTest` instead.

ctk/common/src/main/kotlin/org/codice/compliance/utils/report/Report.kt
```diff
@@ -0,0 +1,241 @@
...
+
+object Report {
+
+    internal const val REPORT_FILE = "report.txt"
+    private const val NOT_TESTED_LIST =
+            "https://github.com/codice/saml-conformance/blob/master/ctk/idp/NotTested.md"
+
+    private val emptyMap: MutableMap<Section, MutableSet<SAMLComplianceException>?> =
+            Section.values().map {
+                it to null
```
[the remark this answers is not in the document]
blen-desta: I had it as null to be able to support skipping a section. Otherwise, if a section is not run, it will mark it as passed which I thought would be misleading.

ctk/common/src/main/kotlin/org/codice/compliance/utils/report/Report.kt
```diff
@@ -0,0 +1,246 @@
...
+            }.toMap().toMutableMap()
```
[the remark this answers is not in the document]
blen-desta: The `.toMap()` is necessary otherwise it says `Unresolvable reference`

ctk/common/src/main/kotlin/org/codice/compliance/utils/report/Report.kt
```diff
@@ -0,0 +1,246 @@
...
+
+    private val exceptionMessages = mutableMapOf<Section, MutableSet<SAMLComplianceException>?>()
+            .apply {
+                putAll(emptyMap)
+            }
+
+    private var currentTestExceptionMessages = mutableSetOf<SAMLComplianceException>()
+
+    private val reportQuietly by lazy {
+        System.getProperty(QUIET_MODE) == "true"
```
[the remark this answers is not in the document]
blen-desta: It's `QUIET_MODE.toBoolean()` in Kotlin. I'll change it to that.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/NameIDPolicyVerifier.kt
blen-desta: I'm not sure but I put it back to how it was

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/NameIDPolicyVerifier.kt
blen-desta: Also reverted this

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/NameIDPolicyVerifier.kt
blen-desta: ☝️ put back as is

blen-desta:
@coyotesqrl @brjeter I changed it so that every SAMLComplainceException will know what section it refers to. That way, we won't need to Sections around.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/CommonDataTypeVerifier.kt
```diff
@@ -38,63 +42,129 @@ class CommonDataTypeVerifier {
...
+                return
+            }
+
+            if (errorCode != null) {
+                Report.addExceptionMessage(section,
+                        SAMLComplianceException.create(errorCode, SAMLCore_1_3_1_a,
+                                message = errorMessage))
+            } else {
+                Report.addExceptionMessage(section, SAMLComplianceException.create(SAMLCore_1_3_1_a,
+                        message = errorMessage))
```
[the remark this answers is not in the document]
blen-desta: I think these new changes address this issues. https://github.com/codice/saml-conformance/pull/211#issuecomment-451630490

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
```diff
@@ -178,25 +223,32 @@ class SubjectComparisonVerifier(private val samlResponseDom: Node) {
...
-            if (format1 != null && format1 != format2)
-                throw SAMLComplianceException.create(SAMLCore_3_3_4_b,
-                        message = "One of the identifier's Format attribute " +
-                                "[${format2 ?: UNSPECIFIED_URI}] is not identical to the " +
-                                "AuthnRequest's Subject identifier's Format attribute " +
-                                "[$format1].",
-                        node = id2)
+            if (format1 != null && format1 != format2) {
+                Report.addExceptionMessage(SAMLComplianceException.create(SAMLCore_3_3_4_b,
+                        message = """One of the identifier's Format attribute
```
brjeter: ✏️ Here's another usage of a raw string.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
```diff
@@ -211,22 +263,25 @@ class SubjectComparisonVerifier(private val samlResponseDom: Node) {
...
-        ) throw SAMLComplianceException.create(SAMLCore_3_3_4_c,
-                SAMLCore_3_4_1_4_b,
-                message = "One of the Response's Subjects had no SubjectConfirmation Methods " +
-                        "[$responseMethods] matching the AuthnRequest's SubjectConfirmation " +
-                        "Methods [$requestMethods].",
-                node = samlResponseDom)
+        if (requestMethods.none { reqMethod -> responseMethods.contains(reqMethod) }) {
+            Report.addExceptionMessage(SAMLComplianceException.create(SAMLCore_3_4_1_4_b,
+                    SAMLCore_3_3_4_c,
+                    message = """One of the Response's Subjects had no SubjectConfirmation Methods
```
brjeter: ✏️ Here's another usage of a raw string.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/SubjectComparisonVerifier.kt
```diff
@@ -211,22 +263,25 @@ class SubjectComparisonVerifier(private val samlResponseDom: Node) {
...
+                            [$responseMethods] matching the AuthnRequest's SubjectConfirmation
+                            Methods [$requestMethods].""",
+                    node = samlResponseDom))
+
+            Report.addExceptionMessage(SAMLComplianceException.create(SAMLCore_3_3_4_c,
+                    message = """One of the Response's Subjects had no SubjectConfirmation Methods
```
brjeter: ✏️ Here's another usage of a raw string.

ctk/common/src/main/kotlin/org/codice/compliance/verification/core/responses/CoreAuthnRequestProtocolVerifier.kt
```diff
@@ -49,56 +50,72 @@ class CoreAuthnRequestProtocolVerifier(
...
+     * 3.3.2.2.1 Element <RequestedAuthnContext>
+     * Verify responses of AuthnRequests With AuthnContext
+     */
     fun verifyAuthnContextClassRef() {
         samlResponse.recursiveChildren("AuthnContext")
                 .flatMap { it.children("AuthnContextClassRef") }
                 .firstOrNull { !ddfAuthnContextList.contains(it.textContent) }?.let {
-                    throw SAMLComplianceException.create(SAMLCore_3_3_2_2_1_a,
+                    Report.addExceptionMessage(SAMLComplianceException.create(SAMLCore_3_3_2_2_1_a,
                             message = """An <AuthnContextClassRef> that is not part of the
```
brjeter: ✏️ Here's another usage of a raw string, outside the scope of your PR which is a little concerning.

brjeter: Did some searching and found a handful of other places where we shouldn't be using the raw string. I can fix this one  and those in a separate PR.

library/src/main/kotlin/org/codice/compliance/report/Report.kt
```diff
@@ -0,0 +1,264 @@
...
+    private var hasExceptions = false
+
+    /**
+     * Adds an exception to the {@code exceptionMessages} for the given {@param exception}'s section
+     *
+     * @param exception - the SAML Compliance Exception to add
+     */
+    fun addExceptionMessage(exception: SAMLComplianceException): Report {
+        val section = exception.section
+        if (!section.isStarted()) {
```
coyotesqrl: ✏️ I believe you can collapse these two overloaded methods into one (`exception` has to be declared first in the signature to support this):
```kotlin
fun addExceptionMessage(exception: SAMLComplianceException, section: Section = exception.section): Report {
        if (!section.isStarted()) {
            exceptionMessages[section] = mutableSetOf(exception)
        } else {
            exceptionMessages[section]?.add(exception)
        }

        currentTestExceptionMessages.add(exception)
        hasExceptions = true
        return this
    }
```
Even if that doesn't work, you can definitely invoke the overloaded method after you've extracted the `section` here.

library/src/main/kotlin/org/codice/compliance/report/Report.kt
```diff
@@ -0,0 +1,264 @@
...
+                    exc.let {
+                        writer.println(Ansi.ansi().fgMagenta().a(it.message).reset())
+                    }
+                }
+            }
+        }
+        writer.println()
+    }
+
+    enum class Section(val title: String, val level: Int = 1) {
```
coyotesqrl: ✏️ Probably better to extract as a top-level enum class at this point since it's shared.

ctk/common/src/test/kotlin/org/codice/compilance/ReportListener.kt
```diff
@@ -0,0 +1,22 @@
...
+import io.kotlintest.TestResult
+import io.kotlintest.extensions.TestListener
+import org.codice.compliance.report.Report
+
+/**
+ * Listener used to reset the Report map for testing purposes before each test
+ */
+object ReportListener : TestListener {
+
+    override fun afterTest(description: Description, result: TestResult) {
```
coyotesqrl: ✏️ Overriding `afterTest` but the docs say "before each test". I know you'll get the same effect, but probably better to get the docs and implementation sync'd. I'd change the docs to match.

ctk/common/src/test/kotlin/org/codice/compilance/verification/core/requests/CoreLogoutRequestProtocolVerifierSpec.kt
```diff
@@ -23,6 +27,8 @@ import java.time.Instant
 import java.util.UUID
 
 class CoreLogoutRequestProtocolVerifierSpec : StringSpec() {
+    override fun listeners(): List<TestListener> = listOf(ReportListener)
```
coyotesqrl: ❓ I keep thinking there should be a better way of doing this instead of overriding the function on every one of our reportable tests, but short of inserting a new base class between `StringSpec` and all of ours, I really can't. And that seems like overkill.

ctk/common/src/test/kotlin/org/codice/compilance/verification/core/responses/CoreAuthnRequestProtocolVerifierSpec.kt
```diff
@@ -48,17 +52,18 @@ class CoreAuthnRequestProtocolVerifierSpec : StringSpec() {
...
                 NodeDecorator(Common.buildDom(response(authContext))).let {
-                    shouldThrow<SAMLComplianceException> {
-                        CoreAuthnRequestProtocolVerifier(AuthnRequestBuilder().buildObject(), it)
-                                .verifyAuthnContextClassRef()
-                    }.message?.shouldContain(SAMLCore_3_3_2_2_1_a.message)
+                    CoreAuthnRequestProtocolVerifier(AuthnRequestBuilder().buildObject(), it)
+                            .verifyAuthnContextClassRef()
                 }
+                Report.getExceptionMessages(CORE_3_3).shouldContain(SAMLCore_3_3_2_2_1_a.message)
+                Report.resetExceptionMap()
```
coyotesqrl: ❓ Do you really need this everywhere, since you've added the report listener?

library/src/main/kotlin/org/codice/compliance/report/Report.kt
blen-desta: That will work. We just use the "explicitly give the section" part in like two places

library/src/main/kotlin/org/codice/compliance/report/Report.kt
blen-desta: Makes sense

ctk/common/src/test/kotlin/org/codice/compilance/ReportListener.kt
blen-desta: That's my bad. I changed the name when Jeter suggested it but forgot the comment

ctk/common/src/test/kotlin/org/codice/compilance/verification/core/requests/CoreLogoutRequestProtocolVerifierSpec.kt
blen-desta: Yeah I tried looking for a better way and other than having a superclass there isn't something I can do

ctk/common/src/test/kotlin/org/codice/compilance/verification/core/responses/CoreAuthnRequestProtocolVerifierSpec.kt
blen-desta: I added this here cause it's in a forAll loop so the listener won't clear it

blen-desta:
test ddf

blen-desta:
build now
test ddf

bakejeyner:
Hero ✔️

blen-desta:
build now
test ddf

library/src/main/kotlin/org/codice/compliance/report/Report.kt
```diff
@@ -134,14 +146,30 @@ object Report {
...
-                writer.print("\t".repeat(section.level))
-                writer.print(section.title)
-
-                // the top level section will not have a status except GENERAL and SCHEMA
-                if (section.level == 2 || section == GENERAL || section == SCHEMA) {
-                    printExceptions(section, writer)
-                } else {
-                    writer.println()
+                when (section) {
+                    CORE_3_3 -> {
```
coyotesqrl: ✏️ I don't like these hard-coded assumptions. Seems better if `section` had an inclusion or exclusion list of profiles so that in the future we could say "section x.y.z only applies to Keycloak", etc.

blen-desta: Is this a good solution? https://github.com/codice/saml-conformance/pull/211/commits/d4c93dacc0bc96eee381440395769ec86a90afe1


This is what will happen now:

**When the DDF profile is NOT run:**
screen shot 2019-01-08 at 8 28 39 am

**When the DDF pprofile is run:**
screen shot 2019-01-08 at 8 29 16 am

blen-desta: 👌 Ahh that's a good idea

blen-desta: Alright so now it will be printed regardless but gray if we're not running the DDF profile: https://github.com/codice/saml-conformance/pull/211/commits/842011d7285cfe58fd58dfd54639cade5919105d

screen shot 2019-01-08 at 9 45 58 am

blen-desta:
build now

library/src/main/kotlin/org/codice/compliance/report/Report.kt
brjeter: 💯

coyotesqrl: 💯

bakejeyner:
Hero ✔️

ctk/idp/src/main/kotlin/org/codice/compliance/web/BaseTest.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import io.kotlintest.extensions.TestListener
+import io.kotlintest.specs.StringSpec
+
+/**
+ * The Base Test class of all tests used to inject the {@link ResultListener} instead of
+ * doing so in each test.
+ */
+open class BaseTest : StringSpec() {
+    override fun listeners(): List<TestListener> = listOf(ResultListener)
```
coyotesqrl: 💯

blen-desta:
build now

